PR #46, opened 2026-06-01
feat(metrics): per-model token buckets for accurate multi-model cost

### LINKED ISSUE #38: metrics: bucket scan_tokens totals by model for correct multi-model cost
## Problem

`crates/metrics/src/scan_tokens.rs` aggregates token totals across all scanned assistant messages (lines ~81–89) but stores only the **last** message's model (line ~96). When a scan range spans a model change (e.g. Sonnet → Opus mid-session), downstream cost math applies one model's per-token rates to the entire token sum, producing an incorrect `costUsd`.

Surfaced by CodeRabbit on #34 (`scan_tokens.rs:97`), classified as a heavy-lift refactor. Deferred from #34's review-fix round because it changes the `ScanResult` schema and intersects the query-time aggregation decision in **ADR 0017**.

## Why deferred

- It is a **schema change**, not a defensive fix: `ScanResult` would need to carry per-model token buckets rather than a single `model` + flat `tokens`.
- It touches the cost pipeline (`compute_cost`) and the `commits.jsonl` record shape, which ADR 0017 deliberately scoped to query-time aggregation. Worth a design pass before implementation.

## Proposed direction (to be confirmed in design)

- Bucket tokens by model within a scan window: `Vec<(model, Tokens)>` or a map.
- `compute_cost` sums per-bucket cost at each model's rates.
- Decide how `commits.jsonl` represents multi-model commits (array of per-model token/cost entries vs. a single blended figure) — coordinate with ADR 0017.

## Acceptance

- A scan spanning two models computes cost as the sum of each model's tokens × its own rates.
- Schema change documented; ADR 0017 updated or a new ADR added if the record shape changes.
- Tests cover a single-model range (unchanged behavior) and a model-transition range.

cameronsjo:
## Summary

Fixes multi-model cost attribution in cadence-metrics: `scan_tokens` previously summed tokens across all messages in a range but stored only the **last** message's model, so `compute_cost` applied one model's rates to the whole sum — wrong whenever a session switches models (opus → sonnet handoffs, fast-mode toggles).

Closes #38

## Changes

- **`ScanResult.by_model`**: per-model token buckets, accumulated via BTreeMap during the scan loop (deterministic order). Existing `tokens` (grand total) and `model` (last message) fields unchanged for backward compat.
- **`compute_cost_by_model`**: total cost = Σ per-bucket (tokens × that model's rates). Unknown models contribute $0, preserving existing behavior.
- **`commits.jsonl` schema**: `costUsd` is now the correct multi-model sum; new `byModel: [{model, tokens, costUsd}]` array, always emitted (single-model ranges produce one entry).

Old rows lack `byModel` — queries treat absence as single-model. ADR 0019 (in the claude-configurations meta-repo) documents the additive schema change.

## Verification

- `make ci` exit 0 (independently re-run by orchestrator in this worktree)
- Metrics crate: 47 tests passing (was 40, +7 new)
- New tests cover: single-model parity (totals unchanged, one bucket), two-model transition (cost = hand-computed per-model sum), unknown-model → $0, multi-bucket commit record shape
- All 40 pre-existing metrics tests pass unmodified

## Summary by CodeRabbit

* **New Features**
  * Commit records now include detailed per-model token buckets and USD cost breakdowns, and scanning reports per-model token totals.

* **Documentation**
  * Clarified that commit cost is computed per model across invocations and that output records include a by-model breakdown array.

* **Tests**
  * Added tests validating per-model token buckets, deterministic ordering, unknown-model handling, and per-model cost aggregation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,6 +74,12 @@ exit 0. They never block a tool call (see [Hook Protocol](#hook-protocol)).
 to add a `_keys` array of raw payload keys to subagent records — useful for
 spotting schema additions across Claude Code releases.
 
+Cost is computed **per model**: when a commit range spans multiple models
+(opus → sonnet handoffs, fast-mode toggles), each model's tokens are priced at
+its own rates and summed. Records carry the breakdown in a `byModel` array
+(`[{model, tokens, costUsd}]`); rows written before this field existed are
+single-model by definition.
+
 ### lab (cadence-lab)
 
 Experimental hooks for the [cadence-lab](https://github.com/cameronsjo/cadence-lab)
--- a/crates/metrics/src/compute_cost.rs
+++ b/crates/metrics/src/compute_cost.rs
@@ -24,21 +24,15 @@ pub fn compute_cost(tokens: &Tokens, model: &str, prices: &Prices) -> f64 {
 
 /// Total cost in USD summed across per-model token buckets.
 ///
-/// Each bucket is billed at its own model's rates. Buckets whose model is
-/// absent from the price table contribute `$0.0` (same behavior as the scalar
-/// `compute_cost`). The result is rounded to 6 decimal places.
+/// Each bucket is billed at its own model's rates via [`compute_cost`], so the
+/// total is the sum of the *rounded* per-bucket costs — exactly reconcilable
+/// with the `byModel` array, which emits those same per-bucket values. Buckets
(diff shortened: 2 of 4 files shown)


cameronsjo:
@coderabbitai review

PR #48, opened 2026-06-01
feat(doctor): subcommand cross-reference + registry export + --quiet (issue #39 P1)

### LINKED ISSUE #39: Graceful version-skew handling when plugin hooks reference subcommands missing from the installed binary
# Graceful version-skew handling when plugin hooks reference subcommands missing from the installed binary

## Summary

When an installed plugin's `hooks.json` references a `cadence-hooks` subcommand that doesn't exist in the locally installed binary, Claude Code surfaces the failure as a mid-session hook error — typically right when the user is in the middle of a tool call (e.g. `gh pr create`). The binary already prints a helpful "upgrade me" nudge, but because it exits non-zero, the nudge is delivered as an *error* rather than a soft warning, and tool execution is interrupted.

The current safeguards don't catch this:

- `run-cadence-hooks.sh` fails open only for a **missing** binary (ADR 0008), not for a **present-but-outdated** binary.
- `cadence-hooks doctor` validates `hooks.json` shell-expansion bugs but does not cross-reference subcommand strings against the binary's own command table.

This issue proposes three layered fixes (one minimal, one preflight, one strategic) plus a release-process decision about the beta channel.

## Repro

1. Install `cadence-hooks-beta` at a version older than `cadence-hooks` stable (this happens whenever the beta channel lags stable — see "Strategic" section below).
2. Install / auto-update a plugin that calls a subcommand only present in the newer stable build (today's example: `cadence nudge-polish-before-pr`, added in `0.10.0`).
3. Trigger the matching tool call — for the example above, `gh pr create *`.

**Observed (debug log excerpt):**

```
[DEBUG] Hook PreToolUse:Bash (PreToolUse) error:
cadence-hooks v0.9.0: unrecognized command or arguments.

This usually means a plugin expects a newer version of cadence-hooks.
  cargo install --git https://github.com/cameronsjo/cadence-hooks.git
  https://github.com/cameronsjo/cadence-hooks/releases/latest

Underlying error: error: unrecognized subcommand 'nudge-polish-before-pr'

Usage: cadence-hooks cadence <COMMAND>
```

Claude Code surfaces this as a hook execution error. The user has to interrupt their workflow, debug, and manually `brew upgrade` before retrying.

## Root-cause analysis

Two independent moving parts:

1. **Plugins auto-update** via the Claude Code marketplace mechanism. A plugin SHA bump can introduce new `hooks.json` entries calling new subcommands.
2. **The binary upgrades manually** via `brew upgrade` (or `cargo install`). Nothing forces it to stay in sync with what plugins now demand.

Both are normal — the gap between them is what bites. The binary today does the right thing semantically (clear error, upgrade pointer) but the wrong thing operationally (non-zero exit → Claude Code treats it as a hook failure).

A secondary cause in the case that triggered this issue: the **beta channel was behind stable**. `cadence-hooks-beta 0.9.0-beta.10` was the latest beta tag, while stable had moved to `0.10.0` and added the new subcommand. A beta user who would normally be "ahead" was unexpectedly "behind."

### P0 — Soft-fail unknown-subcommand in `run-cadence-hooks.sh`

The wrapper script already handles one fail-open case (binary missing). Extend it to a second case (binary present but doesn't recognize the subcommand). The binary's own error message is already user-friendly — preserve it as stderr, but exit 0 so Claude Code treats it as a non-blocking notice.

```bash
#!/usr/bin/env bash
# Dispatch to cadence-hooks binary on PATH. Fails open if not found (ADR 0008)
# or if the installed binary doesn't recognize the requested subcommand (ADR-NEW).
set -euo pipefail

BINARY=$(command -v cadence-hooks 2>/dev/null || true)
if [ -z "$BINARY" ]; then
  exit 0
fi

# Capture stderr so we can detect version-skew errors without blocking tool flow.
err_file=$(mktemp)
trap 'rm -f "$err_file"' EXIT

if "$BINARY" "$@" 2> >(tee "$err_file" >&2); then
  exit 0
fi

ec=$?
if grep -q "unrecognized subcommand\|unrecognized command or arguments" "$err_file"; then
  # Binary already printed its own upgrade nudge to stderr. Fail open.
  exit 0
fi
exit "$ec"
```

**Why this is the right level:** the wrapper script is already where the "binary may be absent" graceful-degradation lives. Treating "binary may be stale" as the same class of problem is consistent. It also means consumers of older plugin versions don't break when they upgrade `cadence-hooks` slower than the marketplace updates plugins.

**Caveat:** This means a user can silently miss a hook firing for an extended period if they ignore the stderr nudge. The P1 preflight closes that gap.

### P1 — Extend `cadence-hooks doctor` to validate subcommand references

`doctor` currently scans `hooks.json` files for shell-expansion bugs. Add a second check: parse every `"${CLAUDE_PLUGIN_ROOT}/hooks/run-cadence-hooks.sh" <namespace> <subcommand>` invocation and cross-reference against the binary's own command table (the same data `cadence-hooks list` emits).

Sketch:

```
cadence-hooks doctor

Scans installed plugin hooks.json files for:
  - shell-expansion bugs (existing)
  - subcommand references not present in this binary (new)
  - subcommand references in a different namespace than declared (new)

Exit codes:
  0  clean
  1  warnings (subcommand mismatches found)
  2  errors (shell-expansion bugs found)
```

Output format should name the plugin, the file path, the line number, and the offending `<namespace> <subcommand>` pair — actionable enough that a user can either `brew upgrade cadence-hooks` or downgrade the plugin without grepping.

This is also the canonical place to *document* which subcommands shipped in which version, useful for the next P2 step.

### P2 — Wire `doctor` into a SessionStart preflight

The cadence plugin's `session-start` script already runs at session start (it injects the `using-cadence` skill context). Have it additionally invoke `cadence-hooks doctor --quiet` and, on warnings, inject a short SessionStart notice into the session: "cadence-hooks 0.9.0 is missing 2 subcommands referenced by installed plugins — run `brew upgrade cadence-hooks` to catch up."

This means the user knows about the skew **before** the next `gh pr create` blows up, not in the middle of the workflow.

Trade-off: SessionStart hooks add a small startup latency. `doctor` is fast (file scan + binary call), so this should be sub-100ms. Make the check opt-out via an env var (`CADENCE_HOOKS_SKIP_PREFLIGHT=1`) for users who'd rather not pay it.

### P3 (strategic, optional) — Beta channel policy

Today's failure was amplified by the beta tap (`cameronsjo/tap/cadence-hooks-beta`) lagging stable. That's the inverse of what a beta channel is supposed to do.

Two options:

**Option A — drop the beta tap.** If the only consumer is the maintainer dogfooding prereleases, `cargo install --git github.com/cameronsjo/cadence-hooks --branch <dev-branch>` covers the use case without the maintenance burden of a second formula. Removes a whole category of skew problems.

**Option B — enforce `beta_version >= stable_version` as release policy.** Every stable release must be preceded (or accompanied) by a beta tag at the same or higher version. CI check on tag push: refuse to publish a `cadence-hooks` stable formula whose version exceeds the most recent `cadence-hooks-beta` formula's version. This keeps the channel useful (early adopters always ahead) without ad-hoc divergence.

Recommend **Option A** unless there's a real second consumer of the beta tap — the cost of maintaining two formulas to provide a "prerelease" channel for a single user isn't paying for itself.

## Acceptance criteria

- [ ] `run-cadence-hooks.sh` fails open with non-zero stderr nudge (not error) when binary returns "unrecognized subcommand"
- [ ] `cadence-hooks doctor` flags subcommand-namespace mismatches against installed `hooks.json` files
- [ ] `cadence-hooks doctor --quiet` returns non-zero only on errors (not warnings), suitable for SessionStart wiring
- [ ] Cadence plugin `session-start` script surfaces doctor warnings into Claude Code session context
- [ ] Decision made on beta channel (kept with policy, or retired with cargo-install dev workflow documented)
- [ ] Release notes for `cadence-hooks 0.11.0` document the new behavior

## Out of scope

- Plugin-manifest-level binary version pinning (`requires_cadence_hooks: ">=0.10.0"` in `plugin.json`). Larger change, requires per-plugin bumps, and only worth doing if a third-party plugin ecosystem materializes. Solo-maintainer-today doesn't pay for the abstraction.
- Auto-upgrade daemon. Too much machinery for a single binary; a SessionStart nudge is sufficient.
- Cargo-installed binary detection (separately tracked in the existing brew/cargo dual-install discussion).

## Notes for the implementer

- The P0 patch must be propagated to every plugin that ships `run-cadence-hooks.sh` — there's no shared library across plugins today; each plugin embeds its own copy. Worth thinking about whether the script should be sourced from the binary itself (`cadence-hooks wrap <namespace> <subcommand>`) to eliminate copy drift.
- `doctor`'s subcommand inventory should be the same data structure that backs `cadence-hooks list`, not a hand-maintained constant. One source of truth.
- The `2> >(tee ... >&2)` redirection in P0 needs verification under `set -euo pipefail` — process substitution exit codes can be surprising. Consider `mktemp` + `2>"$err_file"; cat "$err_file" >&2` instead if portability matters.

cameronsjo:
## Summary

Issue #39 P1: `cadence-hooks doctor` now cross-references every installed plugin's hook commands against the binary's own command registry, so version skew ("plugin references a subcommand this binary doesn't have") surfaces as a diagnostic instead of a mid-workflow hook error.

Closes #39

> **Stacked on #47** (`feat/issue-43-checks`) — merge that first; this PR's base is that branch so the diff stays scoped.

The other #39 layers land elsewhere: P0 wrapper soft-fail is in each plugin repo ([cadence-guardrails#8](https://github.com/cameronsjo/cadence-guardrails/pull/8) et al.), P2 SessionStart preflight goes in the cadence plugin, and P3 (beta tap retirement) is a tap-repo change + issue comment.

### `src/registry.rs` (new)

`HookEntry` + the `HOOKS` catalog move out of main.rs into a shared module — one source of truth backing `list`, hook-name resolution, **and** doctor's cross-reference (the issue's "not a hand-maintained constant" requirement: doctor reads the same data `list` emits, not its own copy). Adds `is_known(namespace, subcommand)` and `namespace_of(subcommand)` lookups.

### `src/doctor.rs`

- **Severity model**: `Error` (shell-expansion bugs, the existing check) vs `Warning` (version skew, new)
- **Invocation parsing**: extracts `(namespace, subcommand)` from both wrapper form (`run-cadence-hooks.sh guardrails guard-gh-write`) and direct binary form (`cadence-hooks rules security-patterns`), any quoting style, trailing flags ignored
- **Cross-reference**: unknown pair → skew warning naming the binary version and remediation; known subcommand under wrong namespace → mismatch warning with the corrected entry
- **Line numbers** on findings (best-effort text search)
- **`--quiet`** flag for SessionStart preflight wiring

### Exit codes

| Mode | Clean | Warnings (skew) | Errors (expansion bugs) |
|---|---|---|---|
| default | 0 | 1 | 2 |
| `--quiet` | 0, silent | **0**, one-line stdout summary | 2, one-line stderr |

`--quiet` returning 0 on warnings is the #39 acceptance criterion that makes it safe to call from a `set -euo pipefail` SessionStart script — the script detects skew by non-empty stdout, not exit code.

## Verification

- `make ci` exit 0 — fmt + clippy (`-D warnings`) + all 20 workspace test suites, **1255 tests passed, 0 failed**
- Binary-crate tests: 56 → 79 (+23)
- Behavioral spot-check against fixture plugin trees (independent of the test suite): clean→0, skew→1 with named pair + version, expansion→2 with line number, namespace mismatch→correct hint, quiet+skew→0 with one-line summary, quiet+clean→0 silent, quiet+errors→2

```
warning [skew-plugin] /tmp/doctor-fix2/skew-plugin/hooks/hooks.json: subcommand 'guardrails guard-future-thing' is not present in this binary (v0.11.0)
  command: "${CLAUDE_PLUGIN_ROOT}/hooks/run-cadence-hooks.sh" guardrails guard-future-thing
  fix: brew upgrade cadence-hooks (or downgrade the plugin)
```

## Behavior change

Shell-expansion findings now exit **2** (was 1) under the new severity contract. The two existing integration tests were updated accordingly.

## Attribution

Implemented by an implementer subagent (TDD, per-cycle commits); independently verified by the orchestrating session (make ci re-run + 7-scenario behavioral check).

## Release Notes

* **New Features**
  * Enhanced `doctor` command now detects shell-expansion bugs and subcommand version mismatches across installed plugins
  * Added `--quiet` mode for streamlined output with clearer exit semantics (0 = clean, 1 = warnings-only, 2 = errors)
  * Improved default scan input detection via manifest support

* **Documentation**
  * Expanded `doctor` command documentation with detailed scanning guidance, behavior contract, and example output


cameronsjo:
@coderabbitai review

PR #49, opened 2026-06-01
fix: mid-line security-pattern line numbers + audit enforcement for cadence-rules migration

cameronsjo:
## Summary

The cadence-hooks side of the cadence-rules binary migration ([cadence-rules#8](https://github.com/cameronsjo/cadence-rules/pull/8)), plus a line-number bug fix surfaced while verifying that migration.

> **Stacked on #48** (`feat/issue-39-doctor`) — the bottom of the stack is #47 → #48 → this. Based here so local `make ci` passes (the cadence-guardrails sibling's hooks.json references subcommands that only exist on the stack).

### 1. `fix`: security-patterns line numbers off by one for mid-line matches

`scan_content` counted `lines()` of the prefix before a match — a partial line ("`data = `" before "`pickle.loads`") counted as a full line, so a line-2 match reported as L3. Found during the cadence-rules migration smoke test. Counts newlines now.

The pre-existing `line_number_accuracy` test masked this by only matching at column 0; the new `line_number_accuracy_for_mid_line_match` test covers the general case (classic tests-codify-bugs).

### 2. `test`: enforce cadence-rules as a binary-dispatch plugin in the audit

Moves `cadence-rules` from `SHELL_PLUGIN_DIRS` (exempt) to `BINARY_PLUGIN_DIRS`, so the registration audit starts checking its hooks.json — `rules validate-frontmatter` and `rules security-patterns` must stay registered there, and the cross-plugin / if-filter checks now apply.

## Coordination

This PR and **cadence-rules#8 must merge together**: with this change, local `make ci` fails if the cadence-rules sibling checkout still has the shell-script hooks.json. (CI is unaffected — the audit skips when sibling plugin dirs are absent.)

## Verification

- `make ci` exit 0 — **1256 tests passed, 0 failed** (run with the cadence-rules sibling on its migration branch)
- TDD on the line fix: new test failed with `left: 3, right: 2` against the old code, passes after
- All 7 registration-audit tests pass with cadence-rules enforced

### DIFF
--- a/crates/guardrails/src/guard_gh_write.rs
+++ b/crates/guardrails/src/guard_gh_write.rs
@@ -44,6 +44,12 @@ static REPO_SUBCOMMAND: LazyLock<Regex> = LazyLock::new(|| {
 static API_REPOS: LazyLock<Regex> =
     LazyLock::new(|| Regex::new(r"/?repos/([^/]+/[^/ ]+)").expect("pattern should compile"));
 
+static GIST_COMMAND: LazyLock<Regex> =
+    LazyLock::new(|| Regex::new(r"gh\s+gist\s").expect("pattern should compile"));
+
+static REPO_FORK_COMMAND: LazyLock<Regex> =
+    LazyLock::new(|| Regex::new(r"gh\s+repo\s+fork\b").expect("pattern should compile"));
+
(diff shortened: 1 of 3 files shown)


cameronsjo:
@coderabbitai review

PR #84, opened 2026-06-10
fix(guard-gh-write): resolve each chained gh write's target per-segment

cameronsjo:
## What

Wave 1, PR2 of the bug-hunt fixes. Routes `guard-gh-write`'s chained-command path
through the `command_segments` primitive added in #83.

## The root cause

`guard-gh-write` resolved a single target for the whole command — `extract_repo_flag_str`
returned the **first** `-R` it found and `resolve_target_repo` ran once over the
entire string. So a benign first write covered an unowned second:

```
gh pr comment -R cameronsjo/owned 1 && gh repo delete evil/unowned --yes   # slipped — exit 0
```

The first `-R cameronsjo/owned` resolved as owned, the chain was allowed, and the
unowned `gh repo delete` was never checked.

## The fix

The `NoLoops` write-detection tail now iterates `command_segments(command)` and
judges each write segment independently — `extract_repo_flag` / repo-positional /
api-repos resolution run on the **segment**, not the whole command. The first
unowned or unresolvable write segment blocks; a write hidden in `sh -c '…'` is
surfaced too. The resolution match is extracted into a `judge_write_segment`
helper.

Loop handling (the AST `AllTargetsExplicit` / `MissingTargets` / `ParseFailed`
arms), the unconfigured fail-safe, gist/fork allowances, and the structured
`BlockMetadata` payloads are all unchanged.

## Verification

- `make ci`: fmt + clippy clean; `guard_gh_write` suite **101 passed** (5 new).
  The lone local audit failure is the documented sibling-wiring lag, skipped on
  GitHub CI.
- Live binary (`CADENCE_ALLOWED_OWNERS=cameronsjo`): the chained bypass now exits
  2; the pre-fix 0.26.0 binary exits 0. Two owned writes, a read-then-owned-write,
  stay exit 0; an `sh -c`-wrapped unowned write exits 2.

Closes cameronsjo/claude-configurations#67

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,12 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
   sees through `sh -c '…'` wrappers, matches path-form git (`/usr/bin/git`),
   reads quoted flags (`"--force"`), and exempts an alias definition only for its
   own segment instead of the whole command line.
+- **guard-gh-write: chained writes resolved per-segment** (#67). A gh write chain
+  is now judged one command at a time — `gh pr comment -R me/owned … && gh repo
+  delete evil/unowned --yes` no longer slips because the benign first `-R`
+  resolved the whole chain. Each write segment resolves its own target (and a
+  write hidden in `sh -c '…'` is seen); the first unowned/unresolvable write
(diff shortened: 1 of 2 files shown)


cameronsjo:
@coderabbitai review

PR #93, opened 2026-06-10
feat(core): model heredoc bodies, command substitutions, and visible assignments

cameronsjo:
## Summary

Wave 3 keystone. `core::shell` modeled chains and `sh -c` wrappers (waves 1–2) but not three shell expansion forms — producing both a **live false-block** and several **misses** that every guard shared. All three are fixed in core, so every guard inherits through `command_segments`.

- **Heredoc bodies (live P1 false-block).** `split_segments` now strips heredoc bodies (`<<WORD` … `WORD`) before splitting. On 0.28.0, a heredoc body line `see the .env file for config` became a fake `see` command with a `.env` operand and hard-blocked prevent-secret-leaks (verified live: exit 2). A **quoted** delimiter (`<<'EOF'`) drops the body wholesale; an **unquoted** delimiter re-appends body lines that carry `$(`/backtick so their substitutions still surface, while prose lines are dropped. `<<<` here-strings are not heredocs.
- **Command substitutions.** `command_segments` extracts `$(…)` (nested-paren tracked) and `` `…` `` bodies in executed context — outside single quotes, inside double quotes counts, backslash escapes honored — and segments them recursively. `echo $(cat .env)`, `curl -d "$(cat .env)" …`, `echo \`cat .env\``, and `$(< .env)` now surface the read; `echo '$(cat .env)'` and an escaped `` \` `` stay literal.
- **Visible assignments.** A `VAR=value` / `export VAR=value` assignment present in the command resolves later `$VAR`/`${VAR}` (outside single quotes). `OP_CMD=op; $OP_CMD item list` is now seen as `op item list`. A bare `$OP_CMD` with no visible assignment (environment-sourced) stays unresolved — fail open, classifier backstop holds.

## Verification

- TDD: 11 guard-level behavioral tests (heredoc FP, `$()`/backtick reads, visible-assignment op-scan) written first and watched failing on the v0.28.0 base, plus 17 direct `core::shell` unit tests. All green.
- Suites: `core::shell` 96 passed; full workspace libs green (core 551, cadence 577, guardrails 551). `make ci` clean apart from the two documented sibling-checkout audit failures GitHub CI skips.
- Live pre/post repro (`pre` = installed 0.28.0, `post` = this branch; exit codes):

| Case | pre | post |
|---|---|---|
| heredoc prose mentioning `.env` (FP) | **2** | **0** |
| `echo $(cat .env)` | 0 | 2 |
| `curl -d "$(cat .env)" …` | 0 | 2 |
| `echo \`cat .env\`` | 0 | 2 |
| `echo $(rm .env)` (writes) | 0 | 2 |
| `OP_CMD=op; $OP_CMD item list` | 0 | 2 |
| `echo "$(op item list)"` | 0 | 2 |
| `echo '$(cat .env)'` (single-quote literal) | 0 | 0 |
| `$OP_CMD item list` (env var, documented gap) | 0 | 0 |

Closes cameronsjo/claude-configurations#116

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,22 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
 ## [Unreleased]
 
+## [0.29.0]
+
+### Fixed
+
+- **core::shell: model heredoc bodies, command substitutions, and visible
+  assignments** (#116 on claude-configurations). `split_segments` now strips
+  heredoc bodies before splitting, so a heredoc line like `see the .env file`
+  no longer becomes a fake `see` command with a `.env` operand — a real
+  prevent-secret-leaks false-block on 0.28.0. `command_segments` additionally
+  surfaces command substitutions (`$(…)`, backticks) in executed context, so
+  `echo $(cat .env)`, `curl -d "$(cat .env)" …`, and `echo \`cat .env\`` are
+  judged as the reads they are; and resolves visible `VAR=value` assignments,
+  so `OP_CMD=op; $OP_CMD item list` is seen as `op item list`. Single-quoted
+  and escaped forms stay literal; an environment-sourced variable stays
+  unresolved (fail open). Every guard inherits these through `command_segments`.
+
 ## [0.28.0] - 2026-06-10
 
 ### Fixed
--- a/crates/cadence/src/prevent_secret_leaks.rs
+++ b/crates/cadence/src/prevent_secret_leaks.rs
@@ -1139,4 +1139,94 @@ mod tests {
         let result = SecretLeaksGuard.run(&make_bash_input("cd /tmp\nenv"));
         assert_eq!(result.outcome, cadence_hooks_core::Outcome::Nudge);
     }
+
+    // ---------------------------------------------------------------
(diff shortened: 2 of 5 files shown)


cameronsjo:
Follow-up: an adversarial security review of the heredoc stripper found a Critical evasion — when the stripper's heredoc model diverged from bash (exotic delimiter char class, or `<<WORD` inside a double-quoted string), it dropped trailing lines bash executes, so a dangerous command after the heredoc reached no guard. A regression vs pre-PR. Fixed in 9d75792 with the safety invariant *drop a body only when its terminator is found before EOF; otherwise keep the lines verbatim*, plus double-quote suppression in delimiter detection and matched-quote parsing for quoted delimiters. Added 5 evasion-regression tests; all five reviewer triggers now block. Verified live (exit 0→2 on each trigger; both heredoc false-positives stay exit 0).

cameronsjo:
Superseded by #94 — GitHub persistently dropped CI re-trigger events on this branch after the security fix (3 synchronize pushes + a reopen, none fired a run). #94 is the identical tree at the security-fix commit on a fresh branch, which fires CI normally. The security fix and its verification are carried over verbatim.

Issue #69, opened 2026-06-04
run-cadence-hooks fails open silently when the binary is stale or missing


cameronsjo:
## Summary

`run-cadence-hooks` (the dispatch wrapper) shells out to the `cadence-hooks` binary on PATH and **fails open** when it's missing or stale — a reasonable safety default, but the failure is **silent**. A guardrail the user believes is enforcing can be quietly inert after a version drift between the installed binary and the dispatch wrapper.

## Asks

1. **Make the open-fail visible.** Emit a one-line stderr notice when the wrapper can't find or parse the expected binary version, so an inert guardrail surfaces rather than passing silently.
2. **Document the reproduce-a-check-in-isolation workflow** in the README — pipe a PreToolUse JSON on stdin to the subcommand to confirm whether a block is the binary or the wrapper:

   ```bash
   printf '{"hook_event_name":"PreToolUse","tool_name":"Edit","tool_input":{"file_path":"…","old_string":"…","new_string":"…"}}' \
     | cadence-hooks rules validate-frontmatter; echo "rc=$?"
   ```

   This is the fastest way to isolate the binary from the wrapper when diagnosing whether a hook is actually firing.

## Context

Verified against v0.15.2. Triaged out of cameronsjo/claude-configurations#44 (meta holding-pen).

cameronsjo:
📋 Implementer-ready plan for this issue is up as draft PR https://github.com/cameronsjo/cadence-hooks/pull/242 (`docs/plans/2026-07-06-surface-the-invisible.md`). Disposition here: ask 2 (isolation docs) verified already shipped in docs/testing.md; ask 1 (visibility) delegated to the cadence#223 wrapper notice — this issue stays open until that lands (Task A0 reconciles the doc tense). Planned 2026-07-06 (Fable backlog sweep).

cameronsjo:
Both asks are now satisfied — closing.

**Ask 1 (make the fail-open visible):** landed via cameronsjo/cadence#288 (merged 2026-07-07): the wrapper's missing-binary and stale-binary paths now emit a once/day stderr notice (`notify_inert()`, date-keyed marker under `${XDG_STATE_HOME:-$HOME/.local/state}/cadence-hooks/`), exit codes unchanged — fail-open semantics (ADR-0001) preserved, silence removed. Spec + verification: cameronsjo/cadence#223.

[...]

PR #99, opened 2026-06-16
fix(session): stop stale-sweep pruning live peers; baseline drift on declared branch (#69, #70)

cameronsjo:
## Summary

Two coordination bugs in `crates/session/` that share `touch_own` and the `SessionRecord` schema, so they ship together as one PR.

### #69 — stale-sweep pruned live peers / self

- **Register before sweep.** `run_start` now reads/builds/writes its own record (which refreshes its mtime to ~now) *before* calling `sweep_stale`. This structurally eliminates the self-sweep-then-minimal-rebuild path that dropped a quiet session's `intent`/`touching` on `/clear` or compaction. The fail-open-on-write-error now also returns before any peer is swept.
- **Never sweep your own file.** `sweep_stale` takes the caller's session id and skips the matching `.<short-id>.json` file (reusing `find_own`'s suffix match) even when aged — defense-in-depth against a self-sweep. CLI/tests pass `""`.
- **Liveness threshold 10 → 30 min.** The heartbeat fires only on Bash/Edit/Write, so a session in a long read/think phase emits no signal; 30 minutes keeps it out of the sweep's reach far more reliably. (A separate cadence-canon PR widens the heartbeat matcher.)

### #70 — heartbeat absorbed branch drift

- New `SessionRecord.declared_branch` field — the session's drift baseline, distinct from `branch` (last-observed HEAD). The commit-time warning (`run_drift`) now compares live HEAD against `declared_branch`.
- `declared_branch` moves **only** on THIS session's own `git checkout`/`switch`: the triggering command is threaded through the heartbeat and classified with the existing `guard::is_branch_switch` (quotes stripped first, matching the sibling guard/drift callers). A peer moving shared HEAD refreshes `branch` but leaves the baseline, so the divergence stays detectable at `git commit`.
- Back-fill rule: a new session sets `declared_branch` = live HEAD; a re-register back-fills it only when absent; a self-switch heartbeat re-baselines it.

### No migration

There is no `deny_unknown_fields`, so pre-upgrade records parse with `declared_branch=None` (drift fails open to `allow()`) and self-heal on the next `session start` back-fill.

## Tests

- Reframed `heartbeat_refreshes_existing_record` (which codified the buggy absorb-on-heartbeat behavior) → `heartbeat_non_switch_updates_observed_branch_but_not_baseline`.
- New: `heartbeat_self_checkout_updates_declared_baseline`, `sweep_never_removes_own_file_even_when_aged`, `start::self_sweep_on_restart_preserves_intent_and_touching`, `branch_drift::peer_checkout_drift_is_detected`, `branch_drift::self_checkout_is_not_falsely_flagged`.
- Signature updates threaded through existing `touch_own_*`, `sweep_*`, `heartbeat_*`, and `seed_record`; the `DEFAULT_STALE_MINUTES` pin is now `30`.

`cargo test -p cadence-hooks-session`: **104 passed, 0 failed.** Full workspace suite green; `cargo build --release` and `cargo clippy --workspace --all-targets -- -D warnings` clean. End-to-end binary smoke confirmed the worked check: peer-moved HEAD nudges (baseline preserved), own switch re-baselines and allows.

Closes cameronsjo/claude-configurations#69
Closes cameronsjo/claude-configurations#70

## Summary by CodeRabbit

* **New Features**
  * Enhanced branch drift detection now compares against a declared baseline.
  * Improved branch switch recognition that filters out prose mentions and heredocs.
  * Better baseline management when you perform branch checkouts.

* **Bug Fixes**
  * Session files properly preserved during restart.
  * More accurate drift detection for multi-tree checkout scenarios.

### DIFF
--- a/crates/session/src/branch_drift.rs
+++ b/crates/session/src/branch_drift.rs
@@ -2,15 +2,20 @@
 //!
 //! Concurrent sessions share one working tree, so HEAD can move out from under
 //! a session when a peer (or the user in another terminal) runs `git checkout`.
-//! The registry already holds the baseline: each session's own `branch` field
-//! is written at `session start` and refreshed by the PostToolUse heartbeat on
-//! every git op. Because the heartbeat keeps `recorded == current` for *this*
-//! session's own checkouts, a mismatch at `git commit` time means HEAD moved
-//! due to something other than this session — "changed out from under you."
+//! The registry holds each session's drift baseline in its `declared_branch`
+//! field: set at `session start` and re-baselined by the PostToolUse heartbeat
+//! *only* when THIS session itself runs `git checkout`/`switch`. A peer moving
+//! shared HEAD refreshes the session's last-observed `branch` but leaves
+//! `declared_branch` untouched — so at `git commit` time a mismatch between
+//! live HEAD and the baseline means HEAD moved due to something other than this
+//! session: "changed out from under you."
 //!
 //! Detection is therefore a pure comparison: live `git branch --show-current`
-//! versus this session's recorded branch. No new state, and no false positives
-//! from one's own branch switches (the heartbeat already caught those).
+//! versus this session's `declared_branch` baseline. No new state on the commit
+//! path, and no false positives from one's own branch switches (the heartbeat
+//! re-baselined those). A record with no baseline yet — a pre-upgrade file that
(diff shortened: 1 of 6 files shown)


cameronsjo:
@coderabbitai review

PR #103, opened 2026-06-17
fix(metrics): atomic subagent writes, current model prices, tail-bounded scan (#94, #95, #96)

cameronsjo:
## What

Three P2 fixes in the metrics loggers (observability — all fail-open, no enforcement impact).

| Issue | Fix |
|---|---|
| #94 | `log-subagent` built its line with `writeln!` (many small writes → torn lines under concurrent appends); now one `write_all`, matching the sibling loggers |
| #95 | `prices.json` lacked `claude-opus-4-8`/`claude-fable-5` (commit cost silently $0 for the default models); added both + an `unpricedModels` record field so an unknown model is loud, not a silent zero |
| #96 | `scan-tokens` parsed the **whole** transcript every commit (paid on commit latency); now byte-scans past the marker and parses only the tail |

Verification: 54 metrics tests pass (existing scan_tokens coverage preserved + 6 new, incl. a marker-substring false-start guard); `clippy --workspace -- -D warnings` clean.

**Scope notes:** #95 prices for opus-4-8/fable-5 come from the `claude-api` skill (input/output); cache tiers are convention-derived (`×1.25`/`×0.10`) — flagged in `_meta`. The existing opus-4-7/4-6 entries ($15/$75) disagree 3× with the skill ($5/$25) and may be logging costs ~3× high — tracked separately as #127 (reinterprets historical data; needs an authoritative price confirmation).

> The CHANGELOG `[Unreleased]` entry will conflict trivially with PR #102 (both append bullets); resolution is union-of-bullets.

Closes cameronsjo/claude-configurations#94
Closes cameronsjo/claude-configurations#95
Closes cameronsjo/claude-configurations#96

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -52,6 +52,15 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
   one per `edits[]` element — that the guards fold over; terminology preserves
   its introduced-vs-existing diff (#63) per edit. Write/single-Edit behavior is
   unchanged.
+- **metrics: atomic subagent log writes, current model prices, tail-bounded
+  transcript scan** (#94, #95, #96 on claude-configurations). `log-subagent`
+  built its JSONL line with `writeln!` (many small writes that tear under
+  concurrent appends) — now one `write_all`, matching the sibling loggers.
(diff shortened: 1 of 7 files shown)


cameronsjo:
@coderabbitai review

cameronsjo:
@coderabbitai review

PR #108, opened 2026-06-20
feat(session): SessionEnd→SessionStart backstop for outro "no loose ends"

cameronsjo:
## What

Adds the **deterministic backstop** for `cadence:outro`'s "no loose ends" contract (claude-configurations#123). outro accounts for every thread a session opens against five terminal dispositions (Done / Checkpointed / Carried / Filed / Declined), but that contract is skill prose — a session can end without `/outro`, or skip its accounting, and work slips away silently. This is the belt to that probabilistic suspenders.

Two new `session` subcommands:

- **`backstop-record`** — a SessionEnd `Logger` (mirrors `session end`) that probes the repo for loose ends (uncommitted/untracked via `git status --short`, unpushed commits via `@{u}..` when an upstream exists, and stashes) and, when any remain, stashes a counts-only marker in the git-excluded `.claude/sessions/`. Fire-and-forget; gated strictly on `hook_event_name == "SessionEnd"`.
- **`backstop-warn`** — a SessionStart `Check` (mirrors the `session start` disclosure) that reads the marker at the next open, emits a one-shot nudge summarizing what was left, and deletes it. **Never blocks** (ADR-0001) — a nudge, exit 0.

## Why deferred (SessionEnd → next SessionStart)

`HookEvent` has no `SessionEnd` variant and `Stop` fires after every assistant turn, so the next SessionStart is the only reliable user-visible end-of-session surface. If outro (or any commit/push) resolved the work, git is clean → no marker → silent automatically. For deliberately-left work, set `CADENCE_NO_OUTRO_BACKSTOP`. The message states only what is observable ("loose ends remained at session end") and **never claims outro was skipped** — outro invocation isn't detectable.

## Notes

- The marker filename carries **no `.json` extension** so the registry's `sweep_stale` never reaps it before the repo is reopened (a `.json` marker would be swept past the staleness window — exactly the "reopened later" case the design serves).
- cadence-canon wiring lands in a companion PR (cameronsjo/cadence-canon). `run-cadence-hooks.sh` fails open on an unrecognized subcommand, so release order doesn't matter.

## Test plan

- `cargo test -p cadence-hooks-session backstop` — 19 unit tests (event-gating, no-signals→no-marker, signals→marker, suppression, warn one-shot/idempotent, render sanitization).
- `cargo test registry_matches_clap_dispatch` + `only_loggers_have_no_event` — wiring invariants green.
- `make ci` green (fmt-check + clippy `-D warnings` + all tests).
- Manual end-to-end against a dirty repo: SessionEnd → marker (2 uncommitted, 1 unpushed on main, 1 stash); PostToolUse → no marker; `CADENCE_NO_OUTRO_BACKSTOP=1` → no marker; SessionStart → `additionalContext` nudge + marker deleted; rerun → silent; clean repo → silent both ways; marker invisible to `git status`.

Closes cameronsjo/claude-configurations#123

## Summary by CodeRabbit

* **New Features**
  * Added a "no loose ends" session backstop that detects uncommitted changes, unpushed commits, and stashed work at session end, then warns you on session start with a non-blocking nudge summarizing what remains.

* **Chores**
  * Version bumped to 0.32.0

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,8 +18,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
   suspenders. `backstop-record` is a SessionEnd Logger (mirrors `session end`)
   that probes the repo for loose ends — uncommitted/untracked
   (`git status --short`), unpushed commits (`@{u}..`, only when an upstream
-  exists), and stashes — and, when any remain, stashes a counts-only marker in
-  the git-excluded `.claude/sessions/`. `backstop-warn` is a SessionStart Check
+  exists), and stashes — and, when any remain *and no live peer is still in the
+  checkout*, stashes a counts-only marker in the git-excluded `.claude/sessions/`.
+  The lights-out gate (only the last session out records) keeps a shared
+  multi-session checkout from recording a live peer's in-progress work as loose
+  ends. `backstop-warn` is a SessionStart Check
   (mirrors the `session start` disclosure) that reads the marker at the next
   open, emits a one-shot nudge summarizing what was left, and deletes it. It
   **never blocks** (ADR-0001) — a nudge, exit 0. The design is deferred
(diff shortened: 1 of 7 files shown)


cameronsjo:
@coderabbitai review

Issue #225, opened 2026-07-05
feat(cadence-hooks): first-class meta-repo / nested-repo awareness — resolve guards against the effective target repo


cameronsjo:
## Problem

cadence-hooks guards resolve the repo they judge against from the **hook payload `cwd`** (the session's working directory), but the *actual* repo a mutation targets is frequently a different one. This is the steady state of a common workflow, not an edge case: sessions open with cwd at a **meta-repo** (itself a git repo on `main`, worktree-enforced) while the work happens in a **different git repo** — nested inside it (a plugin subdir with its own `.git`, gitignored by the parent) or a sibling/worktree.

When the effective target repo ≠ cwd repo, guards misfire and misreport:

- `enforce-worktree` blocks a `cd <target-repo> && git commit` as if it mutated the cwd's branch-mode repo, naming the wrong repo (#213, #224).
- The block/nudge prose offers remedies scoped to the wrong repo (create a worktree / `CADENCE_ALLOW_MAIN` / dismiss), none acknowledging that the command already targets a different repo.

## What already works (and the gap)

Target-repo resolution is a solved problem in this codebase — just not applied uniformly:

- **Edit/Write** → `git_dir_for_input` (`warn_main_branch.rs`) already walks up from the edited file to the nearest `.git`, so file mutations in a nested/sibling repo are attributed correctly (the #26 fix).
- **`git -C <path>`** → `enforce_worktree::git_commit_targets` already resolves the `-C` redirect.
- **`cd <path> &&` prefix** → `parse_work_dir` (`core/shell.rs`, used by `polish_marker_present`) already parses this — but the commit-target resolver does **not** use it, so `cd <target> && git commit` falls back to cwd.

Three resolution mechanisms exist; the Bash `git commit` path uses only two of them.

## Proposal

Make "resolve against the effective target repo" a first-class, shared concern across guards:

1. **Unify target resolution** — a guard's repo decision should key off the effective target (the edited file's repo, or the command's effective cwd after `cd`, or `-C`), not the raw payload cwd. Threading `parse_work_dir` into `git_commit_targets` closes #213/#224 mechanically.
2. **Acknowledge cross-repo targets in messaging** — when the effective target repo ≠ cwd repo, either don't fire (the target repo carries its own policy) or say so explicitly ("this commit targets `<target-repo>`, not `<cwd-repo>`") instead of naming the cwd repo.
[...]

Issue #164, opened 2026-07-01
refactor: unified path/state classification for the git/branch guard family


cameronsjo:
## Problem

The git/branch guard family (`warn-main-branch`, `check-idle-return`, `guard-gh-write`, `warn-subagent-worktree`) has produced 7+ false-positive incidents, each fixed with a one-off carve-out — and each fix has been followed by a sibling misfire in the same family: #151 (stale repo-global idle marker), #152 (path-component carve-outs without canonicalization), #155 (missing unrelated-branch coverage), #158 (read-only `gh` calls blocked inside a loop), plus closed #33/#35. The pattern says the carve-outs are treating symptoms of one shared gap: there is no single, tested notion of "what kind of path/repo/branch/operation is this?"

## Suggested Fix

Extract a shared classification module the whole family consumes:

- **Input:** cwd, path arguments, repo state (branch, worktree kind, default branch), session context (session id, subagent-ness).
- **Output:** a small struct — path class (docs/plans, scratch/gitignored, memory, vault, source), worktree kind (main checkout / linked worktree / agent worktree), branch policy (feature / main / trunk-only project), operation kind (read vs write, local vs remote).
- Guards become thin policy over the classifier; every carve-out lives exactly once.
- Property-test the classifier directly — the misfire history above is the seed corpus for regression cases.

Confidence: likely approach — the refactor cost is real, but the one-off-carve-out treadmill has a demonstrated recurrence rate.

Refs: #151, #152, #155, #158, #33, #35.

cameronsjo:
## Recommendation: incremental extraction, done in a dedicated session

Spike outcome — the "incremental, not big-bang" steer is confirmed, with the evidence stronger than the issue implies: `core::paths` already holds `find_git_root` and `resolve_git_common_dir` (the latter's doc comment already names itself the future primitive for this classifier). A unified classifier is the **capstone that composes existing pure primitives**, not a ground-up rewrite. This should be **documented now and built in a focused follow-up session**, not rushed alongside other work — these are safety/security guards where characterization-tests-first matters more than speed, and each guard migration wants to be its own small, individually-reviewable PR.

### Duplication map (the evidence)

[...]

Issue #153, opened 2026-07-01
discovery: unify per-repo guard config into a single .claude/cadence.json


cameronsjo:
## Problem / Motivation

cadence-hooks now ships **two** per-repo config files that have independently converged on the same shape:

- `.claude/redaction.json` — softens the `redact-external-content` *nudge* (additional patterns + an allowlist). Dogfooded in the meta-repo (claude-configurations#207).
- `.claude/terminology.json` — softens the `terminology` *block* (per-file/term exemptions). Shipped in #121.

Both are: discovered by walking up for the git root (now the shared `core::paths::find_git_root`), fail-open (ADR-0001), serde `camelCase`, read from `<git-root>/.claude/`. As more guards gain per-repo knobs, the count of top-level `.claude/*.json` files grows — more files to discover, document, and reason about. A single namespaced **`.claude/cadence.json`** may be the better endpoint, and it's cheaper to converge now (two files) than later (N).

## Current State

| File | Guard | Softens | Shape |
|---|---|---|---|
| `.claude/redaction.json` | `redact-external-content` | a nudge | `additionalPatterns[]`, `allowlist[]` |
| `.claude/terminology.json` | `terminology` | a block | `exemptions[]` (`paths`, `terms`, `mode`) |

Shared today: git-root discovery (`core::paths::find_git_root`), fail-open defaults, `camelCase`, `<git-root>/.claude/` location.

## Proposal Sketch (to explore, not commit)

A single `.claude/cadence.json` with per-guard sections, each guard reading its own slice; an absent section = today's empty default:

```jsonc
{
  "terminology": { "exemptions": [ /* ... */ ] },
  "redaction":   { "additionalPatterns": [ /* ... */ ], "allowlist": [ /* ... */ ] }
}
```

## Open Questions (the discovery)

- **Back-compat.** Existing `.claude/redaction.json` files live in the wild (the meta-repo dogfoods one, claude-configurations#207). Read both and merge? Precedence if both exist? Deprecate-and-migrate, or support standalone files indefinitely?
- **Loader shape.** Factor a shared `core::config::load_cadence_config(git_root)` that each guard slices, or keep per-guard loaders that each read one file?
- **Versioned envelope.** Does a unified file invite `{ "version": 1, ... }` so the schema can evolve without breaking older binaries?
[...]

PR #226, opened 2026-07-06
fix(guardrails): resolve cd-prefixed commit targets in enforce-worktree

### LINKED ISSUE #213: feedback(cadence-hooks): enforce-worktree blocks `cd <worktree> && git` but allows identical `git -C <worktree>`
## Problem

`guardrails enforce-worktree` blocks a `cd <worktree> && git <mutate>` command run from a primary checkout, but **allows the identical-intent `git -C <worktree> <mutate>`** form. Both target the same worktree and never mutate the primary tree, yet only the `cd` form is blocked. The guard keys its decision on `cwd` and does not resolve a `cd <path> &&` prefix to the command's effective target, so it treats a worktree-targeted commit as a primary-checkout mutation.

This tripped several times in one session: every `cd /tmp/wt-... && git commit` was blocked and had to be rewritten as `git -C /tmp/wt-... commit` to proceed — same operation, same target, just a syntax the guard happens to parse.

## Reproduction

`cadence-hooks 0.45.0`. Both payloads have `cwd` = the primary checkout `/Users/cameronsjo/Projects/claude-configurations`; both target the worktree `/tmp/wt-slate-gates`:

```
# cd-form — BLOCKED (exit 2)
cadence-hooks try guardrails enforce-worktree --payload cd.json
  payload: {"tool_name":"Bash","tool_input":{"command":"cd /tmp/wt-slate-gates && git commit -m x"},"cwd":"<primary-checkout>"}
  Outcome: BLOCK (exit 2)

# -C form — ALLOWED (exit 0)
cadence-hooks try guardrails enforce-worktree --payload c.json
  payload: {"tool_name":"Bash","tool_input":{"command":"git -C /tmp/wt-slate-gates commit -m x"},"cwd":"<primary-checkout>"}
  Outcome: ALLOW (exit 0)
```

## Expected vs actual

- **Expected:** both forms allowed — each targets a worktree, not the primary tree. (Or, if the intent is to block *unresolvable* targets, block both and let `-C` through only when its path resolves to a worktree.)
- **Actual:** `cd`-form blocked, `-C`-form allowed. Inconsistent handling of two spellings of the same operation.

## Notes

Generalizable, not session-specific: any workflow that commits to a sibling/worktree via `cd <path> && git …` from a primary checkout hits this. The `-C` escape hatch works but is non-obvious, and the block's suggested remedies (create a worktree / dismiss / `CADENCE_ALLOW_MAIN`) don't mention that the command *already* targets a worktree. Reporting the seam only — the resolution (parse `cd <path> &&` to the effective target, or align both spellings) is the maintainer's call.

### LINKED ISSUE #224: feedback(cadence-hooks): enforce-worktree ignores cd-prefixed cross-repo targets it already honors via git -C
## Problem

`guardrails enforce-worktree` decides from the payload `cwd` (plus any `git -C` target it parses), but ignores a leading `cd <dir> && …` in the command. With the shell parked in a branch-mode primary checkout, a command that explicitly `cd`s into a *different, sibling* repo before its `git commit` is blocked as if it were mutating the branch-mode repo. The block message names the wrong repo, which makes the failure read as "my commit to repo B is forbidden" when repo B has no such policy.

Hit in a real session: the Claude Code Bash tool's working directory persists between calls, so an earlier command left cwd inside a plugin monorepo (branch-mode, worktree-enforced). The next command — `cd <meta-repo> && git add CLAUDE.md && git commit …`, a legitimate docs commit to the sibling meta-repo — was blocked, with the message naming the monorepo. Workaround was a standalone `cd` first (updating the tracked cwd), after which the identical git command was allowed.

## Evidence

`cadence-hooks 0.47.0`. Reproduced with `cadence-hooks try guardrails enforce-worktree --payload <f>` — layout is a branch-mode repo at `~/repos/meta/mono` nested inside an unenforced sibling at `~/repos/meta`, all payloads with `"cwd": "~/repos/meta/mono"`:

| payload `command` | Outcome |
|---|---|
| `git add CLAUDE.md && git commit -m test` | BLOCK (correct — mutation in the primary checkout) |
| `git -C ~/repos/meta add CLAUDE.md` | ALLOW (`-C` target honored) |
| `cd ~/repos/meta && git add CLAUDE.md && git commit -m test` | BLOCK (misfire — same target repo as the `-C` case) |

So the guard already resolves a cross-repo target for `-C` but not for a `cd` prefix — the two forms name the same repo and diverge.

Secondary observation, possibly its own thing: `try` with `"cwd"` set to the *sibling* repo (not branch-mode, no worktree setup) also returns BLOCK, while the identical live PreToolUse call in that cwd was allowed — so `try` and live may diverge on branch-mode detection or environment. Not dug further.

## Expected vs actual

Expected: a command whose git writes target a repo other than the cwd's branch-mode repo (via `cd <dir> && …`, like via `git -C <dir>`) is judged against the *target* repo. Actual: judged against payload cwd; blocked with a message naming the cwd repo.

cameronsjo:
## Summary

`enforce-worktree` blocked `cd <worktree> && git commit` while allowing the equivalent `git -C <worktree> commit`: the commit-target resolver dropped non-`git` segments, so a `cd`-chain never redirected the judged repo.

- `git_commit_targets` now tracks the cd-effective directory while walking segments, reusing `resolve_cd_target` (made `pub`) plus a new operator-aware `split_segments_with_ops` in `core::shell` so `cd x || …` failure-path cds do not redirect. Commit targets carry a fully-resolved base dir instead of `Option<String>`.
- The Bash arm of `run_enforce` judges each target against its cd-effective directory instead of raw `input.cwd`.
- The block message names the target repo when it differs from where the shell started (detect-and-acknowledge, #224).
- `dismiss-enforce-worktree --repo <path>` keys the snooze marker to the named repo, matching how the block reads it; the default (cwd) behavior is unchanged for existing callers.
- New `docs/adr/0001-fail-open.md` documents the fail-open contract (exit-code rules; a guard's own failure never blocks) and the miss-vs-fail-open boundary — the codebase's most-cited ADR now exists in-repo.

Closes #213
Closes #224

## Scope

The cross-repo `CADENCE_ALLOW_MAIN` facet is deliberately not addressed here — deferred to #164 / #153. #225 (umbrella) stays open; this lands its mechanical portion.

## Testing

- 729 guardrails lib tests green; workspace green; `cargo clippy --all-targets -- -D warnings` clean; fmt clean. CHANGELOG `[Unreleased]` entry included.
- Repro matrix against fixture repos (a primary, its linked worktree, and a second independent primary), payloads piped directly to the built binary (`try` overrides payload cwd with process cwd, so direct invocation):

| payload (cwd = primary) | expected | actual |
|---|---|---|
| `cd <worktree> && git commit` | ALLOW | ALLOW — #213 fixed |
| `git commit` | BLOCK | BLOCK — baseline unchanged |
| `git -C <worktree> commit` | ALLOW | ALLOW — unchanged |
| `cd <other-primary> && git commit` | BLOCK, judged against the target | BLOCK; message names the target repo — #224 fixed |

- `dismiss-enforce-worktree --for 5m --repo <other-primary>` writes the marker under the target repo's git common dir, and the previously blocked payload then passes.
- Reviewer heads-up: running the enforce-worktree Scratch tests from a checkout located under `.claude/worktrees/` (or `/tmp`) produces false results — the scratch fixture paths inherit the checkout's path prefix and trip the `is_claude_managed_dir` / `is_temp_root` carve-outs, so every verdict reads ALLOW. Run the suite from a path outside both. This is a pre-existing property of the test suite's scratch-path construction, not introduced here.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved worktree enforcement so `git commit` is evaluated against the correct target repository when commands use `cd` chains and `-C` redirects (including cross-repo cases).
  * Updated block messaging to clearly indicate when the evaluated target repository differs from the directory the shell started in.
  * Enhanced snoozing: `dismiss-enforce-worktree` now supports `--repo <path>` to snooze enforcement for a specific repository (not just the current directory).


cameronsjo:
## Adversarial review + fixes (two lenses)

A code-review and a security-review pass over this diff each found a guard **bypass** the original push introduced — a PR that fixes a guard-bypass had shipped with a different one inside. Both are now fixed and independently re-verified.

### Finding 1 (Critical) — `||` success-path over-suppression
`cd <primary> || true && git commit` (or `… || exit ; git commit`) run from a worktree: the cd's `next_op != Some("||")` gate assumed the cd fails, but bash's `||`/`&&` are equal-precedence and left-associate, so a *succeeding* cd lands the commit inside `<primary>`. Guard said Allow; real bash commits in the protected repo — the exact case this PR exists to block.

**Fix:** dropped the `||` special-case from the commit-target path entirely — a `cd` segment always updates the tracked effective directory. (`core::shell::parse_work_dir` keeps the heuristic; it only feeds nudge guards — tracked in #229.)

### Finding 2 (regression this PR introduced) — cd option-flag / `cd -` / `$VAR` mis-resolve
The cd handler took the token after `cd` unconditionally as the target, so `cd -P .`, `cd -- .`, `cd -`, and `cd $DIR` from a protected primary resolved to a bogus path → no repo found → **Allow**, where the old cd-blind code **blocked**.

**Fix:** skip leading option-flag tokens to find the real path; if the target is `-`, `$`-prefixed, or absent, keep the pre-cd directory (fail toward blocking) rather than resolving a bogus path.

### Verification (orchestrator, independent)
Re-ran from a checkout outside `.claude/worktrees/` and `/tmp` (both trip the guard's own scratch-path carve-outs):

- `cargo test -p cadence-hooks-guardrails`: **740 passed, 0 failed**; `clippy --all-targets -- -D warnings` clean; `fmt --check` clean.
- Independent probe (hand-built fixtures, payloads piped to a freshly-built binary) — every bypass now Blocks, baselines unchanged:

| payload | verdict |
|---|---|
| `cd <other-primary> \|\| true && git commit` (cwd=worktree) | Block |
| `cd <other-primary> \|\| exit ; git commit` (cwd=worktree) | Block |
| `cd -P . && git commit` (cwd=primary) | Block |
| `cd -- . && git commit` (cwd=primary) | Block |
| `cd $DIR && git commit` (cwd=primary) | Block |
| `cd - && git commit` (cwd=primary) | Block |
| plain `git commit` (cwd=primary) | Block |
| `cd <worktree> && git commit` (cwd=primary) | Allow |
| `git -C <worktree> commit` (cwd=primary) | Allow |

### Follow-ups filed (out of scope here — pre-existing, not worsened by this PR)
- #228 — `sh -c '…git commit'` and quoted-`"git"` wrappers evade commit-target detection (security)
- #229 — `parse_work_dir` shares the finding-1 heuristic; align before any consumer hardens
- #230 — quoted `-C` argument evades detection (`split_whitespace` vs `tokenize`) (security)

PR #245, opened 2026-07-07
fix(guardrails): see through shell wrappers, substitutions & env prefixes in commit detection (#228, #230)

### LINKED ISSUE #228: enforce-worktree: shell-wrapper and quoted-command-word evasions bypass commit-target detection
Found during the adversarial review of #226 (pre-existing — not introduced or worsened there; carved out of that PR's scope deliberately).

`git_commit_targets` in `crates/guardrails/src/enforce_worktree.rs` walks segments with a `split_whitespace` first-token test for `git`, so two whole classes of commands never register as commits at all:

- **Shell wrappers**: `sh -c 'cd /primary && git commit -m x'` — leading token is `sh`, the inner command is never expanded, and the commit runs unguarded. Same for `bash -c`, `zsh -c`, `env git …`.
- **Quoted command word**: `"git" commit -m x` — `split_whitespace` yields `"git"` (with quotes) ≠ `git`; the shell runs it as git.

Both are misses in ADR-0001's miss-vs-fail-open framing: the parser drops a command the shell executes.

Suggested direction from the review: route `git_commit_targets` through `command_segments` + `tokenize` (quote-aware) instead of `split_segments_with_ops` + `split_whitespace` for the git-word test. That would also close the quoted `-C` gap tracked separately.

Refs: #226 review thread, #224.

### LINKED ISSUE #230: enforce-worktree: quoted `-C` argument evades commit-target detection (split_whitespace vs tokenize)
Found during the adversarial review of #226 (pre-existing — same behavior before that PR; not a regression).

`git_commit_targets` (`crates/guardrails/src/enforce_worktree.rs`) computes quote-aware `tokenize(&segment)` to detect a leading `cd`, but then re-tokenizes the same segment with `split_whitespace` for the git/`-C` flag walk. A quoted `-C` argument — `git -C "/some path with spaces" commit -m x` — therefore never parses as a commit target: the whitespace walk sees `"/some`/`path` fragments, the command falls through unguarded, and the shell commits in the `-C` target.

Miss, not fail-open (ADR-0001 framing): the parser drops a command the shell executes.

Fix is likely subsumed by the tokenizer unification proposed in the wrapper-evasion issue (route the git-word/flag walk through the existing quote-aware `tokenize`); filing separately so the gap is tracked even if that refactor is deferred.

Refs: #226 review thread.

cameronsjo:
Closes cameronsjo/cadence-hooks#228
Closes cameronsjo/cadence-hooks#230

## What

`enforce-worktree`'s commit detection missed commits the shell actually executes:

- **Wrapper scripts** — `sh -c 'git commit -m x'`, `bash -c "…"`, `zsh -c '…'`: the commit hid behind the wrapper's leading word.
- **Command substitutions** — `echo "$(git commit -m x)"` / backticks: substitution bodies execute, but were never walked.
- **Environment prefixes** — `env git commit`, `env VAR=x git commit`, and bare `VAR=x git commit`: the prefix word ate the leading-word gate.

All were silent misses on the guard's hard boundary (a commit into a primary checkout reached no guard). Detection now recurses into each segment's *child scripts* via a new `core::shell::child_scripts` seam (wrapper `-c` script, else `$(…)`/backtick bodies — precedence matching `command_segments`), and the prefix-skipper handles `env` plus leading `NAME=value` assignment words.

## Design note: scoped recursion, not a flat `command_segments` routing

The obvious mechanism — iterating `command_segments(command)` instead of `split_segments_with_ops(command)` — is defective for *this* caller, because `git_commit_targets` tracks `effective_dir` across segments. The flat stream splices a child's segments inline, so a subshell's `cd` would move the parent's tracked directory:

```
echo "$(cd /elsewhere)" && git commit -m x     # run from a primary checkout
```

flattens to `[cd /elsewhere, echo …, git commit -m x]` → the commit is judged against `/elsewhere` and **allowed**, while the real commit lands in the primary — a brand-new bypass primitive. Instead, `collect_commit_targets` recurses per child with the directory in effect at that segment: a child inherits the parent cwd at spawn (`cd /wt && sh -c 'git commit'` → `/wt`), but its own `cd`s never leak back out. Child expansion can only *add* commit targets, never move existing ones (widen-only, both directions).

## Attribution: the quoted facets of #230

`"git" commit -m x` and `git -C "/spaced path" commit` were fixed structurally by #241's shared quote-aware `tokenize`. This PR closes #230 by adding the regression tests proving those facets, on top of the wrapper facet fixed here.

## Verification

- 13 new tests (wrapper/substitution/env detection; cd scope isolation both directions; single-quote suppression; nested-wrapper depth bound; documented-miss guards for `env -i`): `cargo test -p cadence-hooks-guardrails enforce_worktree` = 106 passed / 0 failed, zero regressions.
- Binary smoke against a fixture primary checkout (bare subcommand, crafted payloads on stdin): 6 evasion payloads exit 2 (wrapper-cd, wrapper-commit, quoted word, quoted `-C`, env prefix, subshell-cd poisoning attempt), 2 allow controls exit 0.
- `make ci` exit 0 (fmt-check, clippy `-D warnings`, full workspace suite).
- Adversarial security review of the diff ran before merge; verdict recorded as a PR comment.

## Accepted misses (documented, unchanged)

`env -i git commit` / `nice -n 10 git commit` (prefix flags unparsed — miss, never misparse), tokenizer backslash-escape unawareness (deferred by design), substitution inside a `-C` value (pre-existing, fails open), >3 wrapper nesting depth (shares `command_segments`' budget).


cameronsjo:
## Adversarial security review (pre-merge)

A `cadence-forge:security-reviewer` pass ran against the branch diff, framed to find inputs where a `git commit` the shell executes slips past `git_commit_targets` unseen, and to verify the widen-only invariant in both directions.

**Verdict: FINDINGS (3), all fixed in this PR.**

- **Medium — transparent prefix in front of a wrapper was a silent miss.** `child_scripts` tested the raw segment's leading word, while `skip_transparent_prefixes` ran only later in the git arm, so the two transparency mechanisms never composed: `exec sh -c 'git commit'`, `env GIT_AUTHOR_NAME=x bash -c 'git commit'`, `time bash -c 'git commit'` all returned no target. **Fix:** prefix-/assignment-stripping now runs *before* wrapper detection (`child_scripts(argv, segment)` takes the stripped token view). Regression test `transparent_prefix_before_wrapper_composes` (7 forms) + `prefixed_wrapper_cd_still_redirects`.
- **Low — wrapper + trailing substitution was either/or.** `child_scripts` returned the `-c` script *or* substitution bodies, never both, so `bash -c 'true' "$(git commit)"` (the substitution runs in the parent before bash spawns) dropped the commit. **Fix:** `child_scripts` now unions the wrapper script with substitution bodies. Regression test `wrapper_and_trailing_substitution_both_seen`.
- **Low — stale module docstring** listing `sh -c` and `VAR=x git commit` as accepted misses (both now fixed). **Fix:** docstring rewritten to name the *remaining* miss (prefix option-flags like `env -i` / `nice -n`).

**Clean:** widen-only holds both directions — a child's `cd` never re-points the parent's tracked directory (`echo "$(cd /x)" && git commit` still targets parent cwd), and non-commit git under recursion fabricates no target (`$(git rev-parse HEAD)` → empty). `is_assignment_word` can only advance the leading-word start. Making `child_scripts`/`MAX_WRAPPER_DEPTH` `pub` leaks nothing (pure string fns already reachable via `command_segments`).

Post-fix: 109 guardrails tests pass, `make ci` exit 0, binary smoke 9/9 (7 evasions → exit 2 incl. the prefix+wrapper composition, 2 allow controls → exit 0).

cameronsjo:
## Code review (cadence:code-reviewer, quality pass)

CodeRabbit is rate-limited on this PR (33-min reset — its `pass` check is the known green-check-lie, no review ran), so our own reviewer gaggle stands in.

One **Important** finding, now fixed: the module doc comment claimed "only one transparent-prefix or assignment word is stripped ahead of a wrapper" — false, since `skip_transparent_prefixes` is an uncapped `while` loop (and `prefixed_wrapper_cd_still_redirects` proves it with `env exec sh -c …`). Reworded to "any run of … words." Also updated `MAX_WRAPPER_DEPTH`'s doc to name its new second consumer. Both doc-only (`b81f160`).

Otherwise clean: naming/idiom consistent with the file, 14 new tests meaningful and non-overlapping, recursion correctly depth-bounded and cd-scoped in both directions (verified by hand-trace), no dead code or needless allocation.

PR #261, opened 2026-07-08
plan(guard-rm): path-aware rm triage replacing blanket ask rules

cameronsjo:
Cluster-plan draft PR (F) — a path-aware `guard-rm` check replacing blanket recursive-rm `ask` rules that interrupt background sessions on every temp/scratch cleanup.

Key platform fact shaping the design (verified against the current permissions docs): a PreToolUse hook's allow decision cannot bypass settings `ask`/`deny` rules — only an exit-2 block pre-empts (and only over allow rules). So this is a **replacement, not a complement**: the guard becomes the sole discriminator (allow temp roots and managed scratch dirs; block home/vault/repo-root targets; ask the ambiguous middle), paired with an allow rule — and the blanket ask rules retire only after a proving period, so the surface is never unguarded between the two.

Reuse ledger: no new parsing (reuses the existing segment tokenizer), temp-root and managed-dir predicates reused from existing guards, the vault trash-guard as a polarity-inverted template, existing exemption/escape-hatch patterns. Net-new: the check itself plus registry entries.

Mandatory adversarial security review is written into the plan: this is a parser predicate gating an ALLOW decision.

Refs cameronsjo/claude-configurations#361 (epic, cluster F).

### DIFF
--- a/docs/plans/2026-07-08-guard-rm.md
+++ b/docs/plans/2026-07-08-guard-rm.md
@@ -1,0 +1,76 @@
+# guard-rm: path-aware rm triage — implementation plan
+
+> **Plan PR (draft).** Ships as `docs/plans/2026-07-08-guard-rm.md` on `cameronsjo/cadence-hooks`. This document is the whole contract — a fresh implementer session executes it with zero prior context.
+> **Refs (never `Closes` — this is a plan, not the fix):** `Refs` epic `cameronsjo/claude-configurations#361` (joins as cluster F). Not from the blind-spot audit — from live friction: the blanket `ask` rules on recursive rm interrupt Fable/background sessions on every scratchpad wipe and temp cleanup.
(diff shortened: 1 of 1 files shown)


cameronsjo:
## Step-0 gate: GO WITH TRIMS — approved 2026-07-08

Owner-lens panel review; trims are part of the approved scope:

1. **Cut the per-repo exemption config (`.claude/rm-exemptions.json`) from v1** — YAGNI; the temp + managed ALLOW rows cover all described friction. It returns on a real trigger.
2. **Enroll guard-rm as a hooks#164 consumer** — it reuses the exact predicates that refactor consolidates and adds new carve-outs (.git-root, home-child, vault); name those carve-outs as #164 seed corpus so the unification inherits them.
3. **`is_temp_root` is `(repo_root, tmpdir)`** — it tests whether a repo root IS a temp root, not whether an arbitrary rm target lives under a temp dir. Adapt/extract; don't call directly.

Mechanism verified correct against the founding finding (deny/ask rules evaluate regardless of PreToolUse hook output → allow-rule + blocking hook, ask rules retire LAST after the proving period). Security review stays mandatory as task 3.

cameronsjo:
Plan fully executed: binary in #264 (released v0.53.0 — path-aware rm triage through the adversarial security review, Step-0 trims honored: exemption config cut, #164 corpus enrollment, is_temp_root adapted), wiring in cameronsjo/cadence#332 (PreToolUse Bash matcher; prefilter widened to `*rm*` per CodeRabbit review — xargs/tab forms). Ask-rule retirement stays parked for the proving period per the plan. Merging the plan doc.

PR #264, opened 2026-07-08
feat(guardrails): guard-rm — path-aware rm triage (#261)

cameronsjo:
## guard-rm — path-aware `rm` triage

Refs #261.

A path-aware discipline guard for `rm`-family commands. Rather than a blanket
prompt on every recursive delete, `guard-rm` inspects the target paths per
command segment and returns a graduated decision:

- **ALLOW** (silent) — targets under a temp root (`/tmp`, `/private/tmp`,
  `$TMPDIR` incl. `/var/folders` canonicalization) or under a `.claude`-managed
  directory (worktrees, session dirs, scratchpads).
- **BLOCK** (exit 2, disclosed message + escape hatches) — `/`, `~`, `$HOME`,
  a first-level home child, inside `$OBSIDIAN_VAULT`, or a git repo root / any
  path with a `.git` component.
- **ASK** (prompt) — unexpanded vars, command substitution, unresolvable /
  relative-escaping paths, and everything else (the genuine ambiguous middle).

The ALLOW rows are the friction removed; the ASK default means nothing that
prompts today stops prompting unless it is structurally proven safe.

### Charter

A **discipline guard with an allow-granting edge**, not a security boundary.
Fails open (ADR-0001). The catastrophic floor (built-in `rm -rf /` / `~`
circuit breakers plus retained `deny` rules in settings) bounds the
fail-open-plus-allow window. Deliberately NOT in `PROTECTED_GUARDS` — the
disable switch may neuter it.

### Net-new shared-crate capability

Introduces an `Ask` outcome to the core protocol so a guard can emit
`hookSpecificOutput.permissionDecision: "ask"` (exit 0). No prior render path
emitted a permission decision.

### Gate result (ask-over-allow ordering)

The design depends on a hook's `ask` decision overriding a settings `allow`
rule. Confirmed against the official docs: PreToolUse permission decisions
resolve **most-restrictive-first**, in the order `deny`, `defer`, `ask`,
`allow` — `ask` "shows the permission prompt to the user as normal",
overriding an `allow` rule. A live fresh-session smoke test is queued for the
settings posture change (when the broad ask rules retire and this guard becomes
the sole discriminator).

### Segment-aware parsing

Judged per command segment (so `cd /tmp && rm -rf x`, `bash -lc 'rm -rf y'`,
and `$(…)` bodies are each seen), paths resolved against each segment's
effective cwd via the `cd`-chain. Reuses the adversarially-reviewed shell
segmentation primitives — no new recursion.

### #164 seed corpus

This guard's new carve-out classifiers — `.git`-root detection, home-child
detection, and vault-membership — are offered as seed corpus for the
classifier-unification effort (#164).

---

Status: **draft** — building. Decision-table `try` probe table and security
review to follow.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,13 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
+## [Unreleased]
+
+### Added
+
+- **New `guardrails guard-rm` hook — path-aware triage of `rm`-family delete commands (#261).** Instead of a blanket permission prompt on every recursive delete, guard-rm inspects each command's deletion targets and returns a graduated decision: **ALLOW** (silent) for a target under a temp root (`/tmp`, `/private/tmp`, `$TMPDIR`) or strictly under a `.claude/` directory (worktrees, session dirs, scratchpads); **BLOCK** (exit 2, disclosed message) for `/`, `$HOME`, a first-level home child, inside `$OBSIDIAN_VAULT`, or a git repo root / any path with a `.git` component; and **ASK** (a new core outcome — see below) for an unexpanded variable, a command substitution, a `..`-bearing path, or anything else not proven safe (the default). A *discipline guard with an allow-granting edge*, not a security boundary: it fails open (ADR-0001) and is deliberately **not** in `PROTECTED_GUARDS`, so `CADENCE_DISABLE=guard-rm` may neuter it. Verb detection basename-matches the transparent-prefix-stripped leading word against `rm`/`unlink`/`shred`/`truncate` (plus `find … -delete`), so `git rm`, `npm rm`, and look-alikes like `charm` are correctly not deletions. Targets are collected by mirroring `enforce_worktree::collect_commit_targets` — walking segments while tracking the effective cwd through `cd` chains and recursing into `sh -c '…'` wrappers and `$(…)`/backtick bodies with a fresh scope, never flattening via `command_segments` (issue #228). The plugin `Bash`-matcher wiring is a release-gated companion — it merges only once this release ships the subcommand (doctor `hooks-skew` gate).
+- **New `Outcome::Ask` in the core hook protocol.** A guard can now emit a PreToolUse `permissionDecision: "ask"` envelope (exit 0) to force the interactive permission prompt on an operation it can neither prove safe (Allow) nor prove dangerous (Block). Per Claude Code's most-restrictive precedence (`deny` > `defer` > `ask` > `allow`), an `ask` decision prompts the user even when a settings `allow` rule would otherwise auto-approve. `decision_for` maps it to a `"ask"` row in `denials.jsonl`, logged unconditionally (like a hard deny) so a proving period sees guard interventions, not just blocks.
+
 ## [0.52.0] - 2026-07-07
 
 ### Added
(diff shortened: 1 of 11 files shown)


cameronsjo:
## Verification

### Live decision-table probe (binary at `9d28cde`)

Each row pipes a crafted PreToolUse payload to the bare `guardrails guard-rm`
subcommand and classifies the result (exit 2 → BLOCK; exit 0 + `permissionDecision:"ask"` → ASK; exit 0 + empty stdout → ALLOW):

| Verdict | Command |
|---|---|
| ALLOW | `rm -rf /tmp/scratch/foo` |
| ALLOW | `cd /tmp && rm -rf build` |
| ALLOW | `rm -rf <repo>/.claude/worktrees/x` |
| ALLOW | `charm install foo` (not an rm) |
| ALLOW | `git rm --cached secret.txt` (not an fs rm) |
| BLOCK | `rm -rf /` |
| BLOCK | `rm -rf /*` |
| BLOCK | `rm -rf ~` |
| BLOCK | `rm -rf ~/Documents` |
| BLOCK | `bash -lc "rm -rf /"` |
| BLOCK | `rm -rf $VAULT/notes` |
| BLOCK | `rm -rf <repo>/.git` |
| BLOCK | `find ~/Documents -name "*.md" -delete` |
| BLOCK | `find -L ~/Documents -delete` |
| BLOCK | `find -x ~/Documents -delete` (BSD global opt) |
| BLOCK | `find ~/Documents -exec rm -rf {} +` |
| BLOCK | `find ~/Documents -ok rm {} +` |
| BLOCK | `\rm -rf /` (alias-bypass) |
| BLOCK | `rm -rf ~/./Documents` |
| ASK | `rm -rf $BUILD_DIR` (unexpanded var) |
| ASK | `rm -rf /opt/app/cache` (ambiguous middle) |
| ASK | `rm -rf ../../etc` (relative-escaping) |
| ASK | `find -name "*.md" -delete` (no explicit path) |
| ASK | `find -f ~/Documents -delete` (BSD `-f` adds a root) |
| ASK | `find -zzz ~/Documents -delete` (unrecognized option) |

`make ci` (fmt-check + clippy `-D warnings` + full suite) is green from a
carve-out-free checkout; 52 guard-rm unit tests cover every decision-table row.

### Security & code review — adjudicated

An adversarial review framed around "find an `rm` target that should
BLOCK/ASK but resolves to ALLOW" surfaced target-extraction misses (the segment
walk, cwd tracking, and the ask-envelope surface were cleared). All fixed with
regression tests:

- **`find` global options** (`-L`/`-x`/`-D`/…) no longer drop the search root.
  `find_roots` now **never silently defaults to the cwd** — no explicit path, an
  unrecognized leading option, or BSD `-f` all resolve to ASK, so a target the
  parser can't confidently name can never become a silent ALLOW.
- **`find … -exec/-ok rm`** is now detected (previously yielded no target).
- **`\rm`** alias-bypass is now seen as `rm`.
- **`.` path segments** no longer downgrade a BLOCK to ASK.

Documented, mitigated misses (each falls outside a leading-`rm` `allow` rule
too, so it defers rather than auto-approving): `xargs rm`, `sudo rm`, `RM` on a
case-insensitive volume, and `-exec sh -c 'rm …'`.

The per-repo exemption ALLOW row (`.claude/rm-exemptions.json`) is intentionally
deferred from v1 (documented in the module header); it returns on a real
trigger.

PR #273, opened 2026-07-10
fix(core): bound git spawns under a shared per-process deadline + doctor latency scan (#271)

### LINKED ISSUE #271: feedback(cadence-hooks): PreToolUse hooks time out on most runs — ~5s per Bash, ~6s per Edit/Write, guards fail open
## Problem

On a heavy-usage macOS host, cadence PreToolUse hooks are timing out on **most** invocations. Net effect measured across one day of sessions: every `Bash` call carries **~4.7s** of blocking hook wall time on average, every `Edit`/`Write` **~5.9s** — and because Claude Code fails open on hook timeout, the timed-out guards **do not enforce**, so the secret/git protections mostly aren't firing.

## Diagnosis

Health-check scan of the 50 most-recent Claude Code session logs on this machine (all from 2026-07-09), counting per-hook timeout events:

| Hook | Timeouts / runs | Avg wall time |
|---|---|---|
| `prevent-secret-writes` (PreToolUse) | 21/21 | ~5.5s |
| `warn-docs-update` (PreToolUse) | 14/14 | ~10.5s |
| `enforce-worktree` (PreToolUse) | 9/9 | — |
| session guard (PreToolUse) | 9/9 | — |
| `git-safety` (PreToolUse) | 16/24 | — |
| `prevent-secret-leaks` (PreToolUse) | 18/34 | — |
| `capture-insights.sh` (Stop) | 8/8 | — |

- `cadence-hooks --version`: **0.52.0** (Homebrew install, on PATH).
- The plugin-side dispatcher (`run-cadence-hooks.sh`) is a thin `command -v` + exec wrapper — the latency is inside the binary invocation (or subprocesses it spawns), not the shell wiring.
- The pattern is suite-wide (every guard, every wiring plugin), which points at a shared per-invocation cost rather than one guard's logic.
- Possibly relevant environment: sessions often run with a working directory on a OneDrive-synced folder (slow/paged filesystem I/O), and the machine runs many concurrent sessions.

## Root Cause

Unknown. Hypotheses, in rough order of suspicion:

1. Per-invocation cold start or a hung subprocess the binary waits on (a spawned `git`/scan process that stalls until the hook timeout kills it).
2. Filesystem scans (secret-write/leak checks, docs-update detection) walking a large or cloud-synced tree — `warn-docs-update` at ~10.5s is the outlier that suggests scanning breadth.
3. Contention from many concurrent sessions invoking the binary simultaneously.

## Expected vs Actual

Expected: guards complete well inside the hook timeout and enforce on every matching tool call. Actual: most runs hit the timeout, adding ~5–6s of latency per tool call while the guards silently fail open.

cameronsjo:
## What & why

Fixes #271. On a host with slow subprocess I/O (a cloud-synced `.git`, endpoint-protection scanning every exec, heavy concurrent-session load), unbounded `git` probes rode PreToolUse hook processes past their 5s hooks.json budget. Claude Code kills a timed-out hook from the outside and fails open — so the guard suite added ~5s of latency per tool call while **silently not enforcing**, and a killed process can't log its own death. The pattern was suite-wide because every hook is a separate process forking its own `git`, with no caching.

The fix gives each process an **internal deadline** under the external one, so guards abandon a stalled probe in time to decide, exit, and record the degradation themselves.

## What changed

- **`core::deadline`** — a shared per-process budget (`CADENCE_HOOK_DEADLINE_MS`, default 3000ms; `0` disables; positive values clamp into a 1000–4500ms band). The floor stops a tiny value from being a guard-softening primitive; the ceiling stops a huge value from silently neutralizing the mitigation.
- **`core::shell::run_git_bounded`** — a std-only bounded runner (no new deps): stdout drained on a thread (no 64KB pipe deadlock), child killed and reaped at the deadline, `GIT_OPTIONAL_LOCKS=0` on every spawn, tri-state result (`Completed`/`SpawnFailed`/`TimedOut`) so callers distinguish "git answered badly" from "git never answered". Every production git spawn (~14 sites) routes through it.
- **Three fail-closed arms made timeout-aware** — `git-safety` bare-`HEAD` force-push, `guard-push-remote` unresolvable target, `guard-gh-write` unresolvable repo — each gating on **its own** resolution's timeout, never a process-global flag. A genuine resolution failure (git answered, target truly ambiguous) still blocks. `guard-gh-write` resolves the ownership-deciding `origin` before the optional `upstream` refinement so the refinement can't starve it, and a timed-out `upstream` never degrades to origin-only judgment.
- **Loud, two-tier telemetry** — `failopen.jsonl` gains `deadline` (git checks degraded to fail-open) and `deadline_block_suppressed` (a fail-closed block was actually bypassed), both additive (schemaVersion stays 1), plus a stderr breadcrumb. `doctor` warns at ≥3 `deadline` and ≥1 suppressed. Session `heartbeat`/`end`/`backstop-record` migrate to the logged dispatcher so their deadline hits emit.
- **`enforce-worktree` 4-5 → 3 spawns** via a per-invocation `GitProbe` memo; snooze reads now go through the already-resolved common dir (pure filesystem, verified writer/reader anchor parity).

## Adversarial review

Reviewed by a security-reviewer and code-reviewer panel against the built diff (the repo's mandated gate for guard decision-semantics changes). Both cleared the tri-state gating, the origin-before-upstream reorder, the floor clamp, and the memo. Findings fixed in this PR:

- **HIGH (security)** — `guard-push-remote`'s loop arm resolves one probe per looped push (an attacker-controlled count), so a padding flood of bogus remotes could drain the shared budget and time the real ownership probe into a fail-open. Now guarded by an **induced-exhaustion discriminator**: a timeout arriving after ≥24 *completed* probes is volume, not host slowness (a genuinely slow host times out on its first probe), and blocks with a "run pushes individually" message. Regression-gated by an integration test.
- **Important/Low** — drain-handle joined on the abandon path; front-gate probes moved to the tri-state so their timeouts record the sharp reason; `CADENCE_HOOK_DEADLINE_MS` upper clamp; deterministic pre-exhaustion/counter tests.

## `doctor` latency scan (folded in; was planned as a separate PR-B)

The binary can't self-report an external timeout kill (a killed process logs nothing), and a **pure-CPU guard** starved by fork/exec contention never touches the internal deadline — so `failopen.jsonl` misses it too. The one place that degradation is recorded is Claude Code's own session logs, which carry each hook's `durationMs`. `doctor` now scans the recent logs and reports, per cadence subcommand, the count of runs ≥1000ms (healthy is well under 250ms) plus any `hook_cancelled`, with the max observed duration — the table #271's reporter built by hand, automated. Read-only, bounded to the 50 most-recent logs by mtime, PII-free.

**Deviation worth a look on review:** the plan called this a "timeout-**kill** scan," but investigating real session logs (Claude Code v2.1.206) found *no clean kill signal* — a slow-but-completed hook logs `hook_success` with a high `durationMs`, `hook_cancelled` is rare and ambiguous, and nothing records "timeout." So this measures **latency** (the signal that is present and cleanly per-guard) rather than kills, which is both honest about the logs and a more direct read on the actual problem. Folded into this PR rather than stacked as a separate one because it's additive and read-only, and folding avoids the stacked-PR base-deletion hazard.

## Honest scope

The deadline bounds **in-process git time only** — the wrapper/fork-exec slice before `main` runs (the EDR tax) is outside its reach; that residual is the process-fan-out follow-up (single-dispatch consolidation). And on a sustainedly slow host, the ownership arms *will* degrade to fail-open as the normal case — with the `deadline_block_suppressed` telemetry as the backstop that makes it visible.

## Notes

- File overlap to watch on merge: design PRs #267 (enforce-worktree) and #268 (git/branch guard classifier), and branch `feat/claude-config-dir` (touches `metrics/common.rs` + `doctor.rs`).
- A schema-doc companion for the two new `failopen.jsonl` reasons lands in the cadence monorepo separately.
- `make ci` green (fmt + clippy `-D warnings` + 28/28 test binaries) from a carve-out-free worktree.

Closes cameronsjo/cadence-hooks#271


cameronsjo:
@coderabbitai review

cameronsjo:
@coderabbitai review

cameronsjo:
@coderabbitai review

Issue #152, opened 2026-07-01
feedback(cadence-hooks): warn-main-branch carve-outs match path components without canonicalizing ..


cameronsjo:
**Nit, surfaced by two independent reviewers on #119 (the `docs/plans/` carve-out PR).**

Both `is_claude_managed_dir` and the new `is_plan_doc_dir` in `crates/guardrails/src/warn_main_branch.rs` scan `dir.components()` for their marker components (`.claude`, or consecutive `docs`→`plans`) without canonicalizing. `Path::components()` preserves `..` as `ParentDir`, so a crafted `file_path` like `docs/plans/../../src/main.rs` resolves on disk to a real product file yet textually carries the exempting components → the on-`main` nudge is suppressed for a file that should warn. The same "match anywhere in path" breadth also exempts a vendored `.../vendor/lib/docs/plans`.

**Severity: low / advisory.** `warn-main-branch` emits a NUDGE (exit 0 — not a block), is fail-open (ADR-0001), and the `file_path` is authored by the same agent the nudge serves (no adversarial incentive). There is no I/O sink. Both reviewers explicitly ruled it acceptable for #119 — filing here so the shared property is a conscious choice rather than an unexamined gap.

**Options**
1. **Leave as-is** — consistent across both carve-outs; impact is advisory-only.
2. **Reject/strip `..` components before matching** — one fix covers both predicates; closes the evasion.
3. **Anchor to repo-root** — exempt only `<repo>/docs/plans` and `<repo>/.claude`, removing the vendored-path breadth too. More code, and a behavior change for the existing `.claude/` carve-out.

If touched, harden **both** predicates together for consistency. Sibling to claude-configurations#226 (whose carve-out surfaced it).

---
_Migrated from cameronsjo/claude-configurations#230 (consolidating ecosystem issue filing onto cadence-hooks)._

PR #284, opened 2026-07-10
refactor(core): gitstate + pathclass foundation, guard-rm first consumer (#164 PR1)

cameronsjo:
## PR1 of the #164 umbrella — unified path/state classification

The git/branch guard family (`enforce_worktree`, `warn_main_branch`, `warn_subagent_worktree`, `guard_rm`, and the branch-resolver inlines) had **no single, tested notion of "what kind of path / repo / branch is this?"** Repo-root resolution, branch/HEAD resolution, worktree-vs-primary detection, path-class membership, and the lexical `..` normalization every carve-out depends on were each re-derived and inconsistently applied — the root gap behind 7+ false-positive incidents, each patched with a one-off carve-out.

This is the **additive foundation**: two pure `core` modules the family will consume as thin policy in later PRs.

### What lands

- **`core::pathclass`** — generalizes `guard-rm`'s shipped `classify_path`/`TargetClass` prototype into a shared `path → PathClass` map: `Temp`, `ClaudeManaged`, `DocsPlans`, `HomeChild`, `GitRoot`, `Source`. Preserves the **load-bearing ALLOW-before-BLOCK precedence** and folds in #152's lexical `..` normalization so no carve-out ever matches an uncanonicalized path.
- **`core::gitstate::GitState::resolve`** — a **pure filesystem walk** (no `git` spawn) exposing `repo_root`, `git_common_dir`, `branch`, `worktree_kind` (primary/linked), `default_branch`. Built on the existing `paths::find_git_root` + `resolve_git_common_dir`.
- **Facts, not policy.** Neither module exposes an `is_main`/block/allow verdict — each guard keeps its own policy *and* its own fail-direction, so a shared classifier feeding a BLOCK guard (PR4) can never drift one guard's decision into another's.
- **`guard-rm` refactored as the first consumer, byte-identical.** All 53 existing `guard_rm` tests pass unchanged; its root/home/vault policy stays guard-local (D5 defers `memory`/`vault` to their second consumer).
- **Seed-corpus regression suite** locking the incident history (#152/#33/#35) at the classifier level.

### Verification

- `make ci` green (fmt-check + clippy `-D warnings` + all tests).
- **Zero behaviour change**: the 53 `guard_rm` characterization tests are the byte-identical lock; any verdict diff would be a bug. core: 452 → 475 tests (new module + seed corpus); guardrails unchanged verdicts.

### Sequence

PR2–PR5 each depend only on this PR and land individually (PR4, the only BLOCK guard, carries a mandatory security-review gate). See the plan doc on `plan/gitstate-path-classifier-164`.

## Summary by CodeRabbit

* **New Features**
  * Added shared, deterministic path classification covering temp locations, managed directories, docs plans, home child paths, and Git roots.
  * Added filesystem-based Git checkout resolution (repo root, common git dir, branch/detached HEAD, default branch, and linked worktree vs primary).
  * Exposed the new shared classifiers to core and reused them from `guard-rm`.
* **Bug Fixes**
  * Hardened path normalization against `.`/`..`-based bypass attempts and preserved `guard-rm` decision precedence.
* **Documentation**
  * Updated changelog with the new Unreleased entry describing these capabilities.


cameronsjo:
**Added a hardening commit (`2982a1e`): canonicalize `GitState.repo_root` + `git_common_dir`.**

An adversarial pass over the deferred PR4b (enforce onto GitState) found a latent footgun already in this PR's `GitState`: `git_common_dir` is un-normalized. A real `git worktree add` proves it:

```
primary: /var/.../primary/.git
wt:      /private/var/.../primary/.git/worktrees/wt/../..
raw ==   false   |   canonical ==   true
```

Two divergences stack — the `../..` from `resolve_git_common_dir`'s `commondir` join, and a `/var` vs `/private/var` symlink split (`find_git_root` keeps the start path's symlink form; git stored the canonical one). A BLOCK guard scoping "same repo" by comparing common dirs (PR4b's target) would read a primary and its own worktree as *different* repos → silently loosen. `GitState` now canonicalizes both identity paths so they're comparison-safe by construction and match git's `--path-format=absolute`. Regression test locks primary + real worktree → one canonical common dir.

Issue #236, opened 2026-07-06
feat(session): SessionStart worktree-posture disclosure — say the wall exists before the first write hits it


cameronsjo:
## Problem

`enforce-worktree` is the first thing a session hears about worktrees — an exit-2 block at the first Edit/Write in a primary checkout. There is no proactive counterpart: `cadence-hooks session start` already reads the repo and current branch at SessionStart, but only records them for peer disclosure. It never says "this is a branch-mode repo; feature work belongs in a worktree."

The cost of that silence is the #234 shape: `git checkout -b` passes, subprocess tree-mutations (`uv add`, `sed -i`, redirects) pass, and a branch-first session accumulates tracked state in the shared primary tree before the tripwire fires — by which point unwinding means discarding or relocating real work. #232 was adjacent: the guard's target resolution got repo-aware while the session's understanding of where it should be working never got set at entry.

## Suggested Fix

Add a **worktree-posture disclosure** to the SessionStart output (same surface as the peer/lane disclosure):

- When cwd is a **primary checkout of a branch-mode repo** (same detection `enforce_worktree.rs` already uses, minus the mutation trigger): one line, e.g. *"Branch-mode repo, primary checkout: feature work starts in a worktree (`EnterWorktree` / `git worktree add`) — the first Edit/Write here will be blocked."*
- When cwd is already a **linked worktree**: optionally disclose that (cheap confirmation the posture is right).
- Respect the existing exemptions so the line never fires where the wall wouldn't: `CADENCE_ALLOW_MAIN`, `CADENCE_NO_ENFORCE_WORKTREE`, temp-dir repos, an active `dismiss-enforce-worktree` snooze.
- Once per session (it's SessionStart), no new snooze machinery needed.

This converts the guard from tripwire-only to disclosed-then-enforced: the wall still holds, but sessions stop discovering it mid-work. #234's subprocess seam remains real and worth its own fix; this shrinks its window from "whole session until first native write" to "sessions that ignore an explicit entry notice."

## Companion filing

The prose/methodology half (attune/using-worktrees entry steering, draft-PR + continuous-push posture, rules reconciliation) is filed on the plugin tracker — cameronsjo/cadence#278. Related: cameronsjo/cadence-hooks#232 (exemption scoping, closed), cameronsjo/cadence-hooks#234 (subprocess coverage, open), ADR-0030 (worktree-per-session invariant).

## Labels

enhancement, guardrails

cameronsjo:
[...]

Issue #234, opened 2026-07-06
feedback(cadence-hooks): enforce-worktree misses subprocess tree-mutations (uv add / sed -i / redirects) on primary checkout


cameronsjo:
## Problem

`guardrails enforce-worktree` blocks the native authoring path (`Write`/`Edit`,
`git commit`) in a primary checkout of a branch-mode repo, but **allows Bash
subprocess mutations of tracked files** (`uv add`, `sed -i`, `>>` redirects). A
branch-first workflow therefore accumulates real tracked-file state — deps,
lockfiles, generated files — in the shared primary tree *before* the tripwire
ever fires. By the time a `Write` or `git commit` trips it, unwinding is costly
(discard the regenerable changes, delete the branch, recreate it under a
worktree). This is a recurring pattern in practice, not a one-off.

## Environment

- `cadence-hooks 0.47.0`
- Hook: `guardrails enforce-worktree`
- Context: on a feature branch inside a **primary checkout** of a branch-mode
  repo (worktree not yet created).

## Reproduction

Rebuilt real payloads and ran them through the hook
(`cadence-hooks try guardrails enforce-worktree --payload <file>`), `cwd` = a
primary checkout of a branch-mode repo. Decision is the exit code (2 = block,
0 = allow):

| command (`Bash` unless noted) | tracked-file effect | decision |
|---|---|---|
| `uv add markdown-it-py mistune` | writes `pyproject.toml` + `uv.lock` | **ALLOW (exit 0)** |
| `echo 'x=1' >> pyproject.toml` | writes `pyproject.toml` | **ALLOW (exit 0)** |
| `sed -i '' 's/foo/bar/' src/…/config.py` | writes tracked source | **ALLOW (exit 0)** |
| `git checkout -b spike/foo` | pointer only, no mutation | ALLOW (exit 0) — correct |
| `Write …/chunker_lib.py` | writes tracked source | **BLOCK (exit 2)** |
| `git commit -m wip` | commit boundary | **BLOCK (exit 2)** |

The `Write` block message and the two `git commit`/`Write` exit-2 blocks are
correct. The first three rows are the gap: subprocesses that mutate tracked
files in the primary tree pass straight through.

## Expected vs actual

- **Expected:** a subprocess that writes tracked files in a primary checkout is
  feature-work mutation and should block (or at least nudge) the same as
  `Write`/`Edit`.
- **Actual:** only the native file tools and `git commit` block; `uv add`,
[...]

PR #288, opened 2026-07-10
refactor(guardrails): drop enforce-worktree cross-guard borrow (#164 PR4a; PR4b resolution deferred)

cameronsjo:
## PR4a of the #164 umbrella — drop enforce-worktree's cross-guard borrow (safe slice)

PR4 in the plan is "enforce-worktree onto GitState" — the **only BLOCK guard**, flagged highest-risk with a mandatory security gate. Analysis surfaced a real loosening hazard in the risky part, so this PR ships the **safe, zero-decision slice** and defers the resolution rewrite to a gated PR4b.

### What lands here

- Relocate `git_dir_for_input` to `core::worktree` (next to the carve-out predicates already there since #236), so `enforce-worktree` and `warn-main-branch` each import the **single core definition** instead of `enforce-worktree` borrowing `warn-main-branch`'s copy. This is the plan's explicit "drop the cross-guard `pub(crate)` borrows" deliverable.
- Pure relocation + import redirect. **No change to enforce-worktree's block decision or fail direction.**

### "No new miss" reasoning (relocation)

The diff moves `git_dir_for_input` verbatim and redirects three imports. enforce-worktree's `run_enforce` / `assess_dir` / `collect_commit_targets` logic is untouched; its BLOCK policy, fail-open (ADR-0001), and both fail directions are byte-identical. **All 869 guardrails tests pass unchanged**, which is the characterization lock. There is no new shell-parsing or fail-direction change here, so the mandatory adversarial review attaches to PR4b (below), not to this relocation.

### DEFERRED to PR4b (carries the mandatory Opus security-review + "no new miss" artifact)

Routing enforce-worktree's repo-root / common-dir resolution through `GitState`. A concrete hazard makes this a careful, gated change, not a mechanical swap:

> `resolve_git_common_dir` returns an **un-normalized** path for a linked worktree (`…/.git/worktrees/wt/../..`) while the primary's is `…/.git`. The Edit/Write arm compares `target_common == cwd_common` by **string** — so a session in a worktree editing a file in its own primary tree would compare *unequal* → judged "foreign" → **ALLOWED**, where git's canonical form makes them equal → **BLOCKED**. That is the exact ADR-0030 cross-tree collision the guard exists to stop — a silent loosening of the BLOCK guard.

The fix (lexically normalize both common-dirs before comparing) is known and small, but landing a change that can loosen a BLOCK guard belongs with the security gate and a full characterization pass over the enforce suite from a carve-out-free checkout.

Stacked on #284 (base `feat/gitstate-classifier-164`); retarget to `main` once PR1 merges.

## Summary by CodeRabbit

* **Refactor**
  * Consolidated worktree and path-resolution handling into shared logic for greater consistency and maintainability.
  * Existing worktree enforcement and main-branch warning behavior remain unchanged.

* **Documentation**
  * Updated the unreleased changelog to reflect the internal restructuring and clarify that no user-visible behavior has changed.


cameronsjo:
**Correction to this PR's deferred-PR4b note.** The body says the fix is to "lexically normalize both common-dirs before comparing." That is **insufficient** — verified empirically. The two common-dir strings diverge *two* ways: the `../..` un-normalization AND a `/var` vs `/private/var` symlink split. Lexical `..`-resolution leaves the symlink halves unequal (`/var/...` ≠ `/private/var/...`).

The correct fix is `std::fs::canonicalize` (resolves `..` **and** symlinks), which also reproduces git's own `--path-format=absolute` output. It's landed at the foundation in PR1 (#284, commit `2982a1e`): `GitState.repo_root` + `git_common_dir` are now canonicalized, so PR4b consumes clean, comparison-safe facts rather than each consumer having to remember to normalize. This PR (the borrow-drop) is unaffected.

cameronsjo:
The risky part of PR4 — the GitState **resolution** migration on the BLOCK guard — shipped as #293 (re-implemented onto #234's rewritten enforce, with the mandatory security review). This PR's remaining scope is only the orthogonal **borrow-drop** (relocating `git_dir_for_input` to `core::worktree` so enforce stops importing it from `warn_main_branch`) — pure decoupling cleanliness, not correctness. Its branch is stale (pre-#234/#287/0.57.0). Leaving open as an optional low-priority cleanup; rebuild onto current main or close if not worth it.

cameronsjo:
@coderabbitai review

PR #293, opened 2026-07-11
refactor(guardrails): route enforce-worktree onto GitState, spawn-free (#164 PR4b)

### LINKED ISSUE #299: GitState::resolve / find_git_root walks past a nonexistent start path into the enclosing repo
`core::paths::find_git_root` (consumed by `GitState::resolve`) validates only ancestor `.git` existence, not that the *start* path itself exists — so it walks past a nonexistent leaf directory into a real parent repo and resolves *that* repo's state.

**Divergence from the old `git -C <dir>` behavior** (same input, `<repo>/does-not-exist` never created):
- Old: `git -C <repo>/does-not-exist branch --show-current` → exit 128 → `None` (fail-closed).
- New: `GitState::resolve(Path::new("<repo>/does-not-exist"))` → `Some { branch: Some("main"), .. }`.

**Impact today (nudge-only):** `warn-branch-base` (`git checkout -b feat <nonexistent-cwd>`) and `nudge-upgrade-after-push` (`cd <nonexistent> && git push`) can fire a spurious *advisory* nudge for a command whose real `cd` fails and whose guarded action never runs. Surfaced by the #164 PR5 code-review.

**Why it matters more soon:** `enforce-worktree` (a **BLOCK** guard) is migrating onto `GitState` (#164 PR4b). There, the same walk-up would let `cd <nonexistent-subdir-of-primary> && git commit` resolve to the primary and BLOCK a commit the shell never runs. PR4b already guards this with a per-consumer `!dir.exists()` check; the cleaner fix is at the primitive.

**Fix:** `GitState::resolve` should return `None` when `start` does not itself exist (or `find_git_root` should reject a nonexistent start), matching git's `-C` semantics. The enforce Edit/Write arm already ascends to the nearest existing ancestor first, so new-file writes are unaffected.

feedback(cadence-hooks) — surfaced 2026-07-11 by #164 PR5 review. Refs #164.

cameronsjo:
## The final #164 umbrella PR — enforce-worktree onto GitState (the only BLOCK guard)

Re-implemented onto **main's #234-rewritten enforce** (which renamed `collect_targets`, added the subprocess-mutation-nudge subsystem, and added `GitProbe` call sites). A textual merge of the original PR4b onto #234 compile-broke — #234's new code called the old probe methods PR4b removed — so this is a re-implementation, not a rebase.

### The migration (byte-identical by construction)

`GitProbe`'s per-invocation memo swaps its `git rev-parse --show-toplevel` / `--path-format=absolute --git-common-dir` subprocess probes for **one `GitState::resolve` memo per directory**. Crucially, the `common_dir()` / `repo_root()` accessors **keep their `Option<String>` signatures** — `GitState` canonicalizes both to the same `--path-format=absolute` form the old probes returned — so every call site is textually unchanged:

- the Edit/Write same-repo scoping (`target_common == cwd_common`),
- the `git commit` cross-repo guard,
- **#234's subprocess-mutation-nudge channel** (`mutation_nudge`),
- the snooze marker, and the block message.

**enforce-worktree now spawns zero `git`.**

### Also folded in — the #299 primitive fix

`GitState::resolve` returns `None` for a nonexistent `start` (matching git's `-C` failure), so it can't climb lexical ancestors up into the enclosing repo and resolve a target the shell never reaches — which for this BLOCK guard would be a **false block**. This replaces a per-consumer existence guard with a fix at the primitive (and closes the nudge-only exposure PR5's review found on `warn-branch-base`/`nudge-upgrade-after-push`). **Closes #299.**

### #271 upshot — a strict improvement

enforce-worktree no longer spawns `git`, so it is now **immune** to a slow/hanging `.git` rather than merely bounded. Its two `deadline_failopen` cases flip from "times out, records a deadline row" to "resolves instantly, records nothing."

### Verification

- All **130 enforce unit tests** pass; full workspace suite green (serial, from a carve-out-free checkout).
- **Empirical end-to-end** (built binary, real primary checkout, `CADENCE_ALLOW_MAIN` cleared): primary edit → **BLOCK**; worktree edit → ALLOW; cross-tree worktree→primary edit → **BLOCK** (the canon-fix case); `sed -i` in a primary → **NUDGE** (#234's channel, now GitState-resolved).

### Security gate

Mandatory `cadence-forge:security-reviewer` (Opus) pass running now, framed to hunt for *loosening* (allow/nudge-where-block). The "no new miss" artifact will be posted here on completion.

Supersedes the borrow-drop scope of #288 as a separate concern (that guard's `git_dir_for_input` relocation is orthogonal cleanliness, not correctness — left out to keep this PR's diff focused on the resolution the security gate reviews).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved worktree detection for symlinked paths and nonexistent locations.
  * `enforce-worktree` now avoids invoking Git during enforcement, improving reliability when Git is unavailable or unresponsive.
  * Updated deadline fail-open behavior so enforcement completes without unnecessary timeout warnings.

* **Documentation**
  * Updated the changelog with the revised worktree resolution and deadline behavior.


cameronsjo:
## Security review — "no new miss" artifact ✅

An adversarial security-review pass (Opus, framed to hunt **only for loosening** — any input that now ALLOWS where it previously BLOCKED, the one severe failure mode for a BLOCK guard) reviewed the full diff, the guard (`crates/guardrails/src/enforce_worktree.rs`), the resolver (`crates/core/src/gitstate.rs`), and its primitives (`crates/core/src/paths.rs`, `worktree.rs`).

**Conclusion: no new loosening.** Every ALLOW-where-BLOCK vector resolves to parity with the old `git rev-parse` behavior.

| # | Vector | Finding |
|---|---|---|
| 1 | is_primary / repo-identity divergence | `state.is_primary()` is `.git.is_dir()` — **byte-identical** to the old `is_primary_checkout(repo_root)`. A real primary always resolves and always classifies primary; submodules/linked/bare present `.git` as file-or-absent and were never blocked under either impl. No input flips a primary to allow. |
| 2 | Same-repo comparison (`git_common_dir ==`) | Old arm compared two canonical `--path-format=absolute --git-common-dir` outputs; new arm compares two `canonicalize_or(resolve_git_common_dir(...))` — also canonical. The `/var`↔`/private/var` + `../..` divergence lived in the *pure primitive*, not the old spawn path; canonicalization makes the spawn-free path **match** git, closing that gap without opening a new one. The real-git test `primary_and_linked_worktree_share_one_canonical_common_dir` proves the load-bearing equality. |
| 3 | Existence guard (`!dir.exists()` → allow) | Restores exact parity with git's fail-open on a nonexistent dir. Not a bypass primitive: if a commit target doesn't exist at hook time, the real `cd <dir> && git commit` short-circuits (the commit never runs), so allow is safe. |
| 4 | `canonicalize_or` fallback | For the same physical common dir, both raw forms are guaranteed traversable (`resolve_git_common_dir` already confirmed `HEAD` reachable), so both canonicalize to the same string. A one-sided fallback-to-raw requires a sub-millisecond TOCTOU deletion of the primary `.git` mid-resolution — which also breaks the real git op. ADR-0001 fallback; no non-race divergence. |
| 5 | Snooze / bypass attribution | Reader uses `state.git_common_dir` (canonical); dismiss writer anchors the marker under `git rev-parse --path-format=absolute --git-common-dir` (also canonical). Both resolve to the same marker path — no stale-snooze honoring, no active-snooze miss. |

### Nits (no exploit path — hardening notes only)

- The point-4 `canonicalize_or` one-sided-fallback race could *theoretically* loosen scoping, but only under a self-defeating TOCTOU (deleting the primary `.git` mid-resolution also breaks the operation). Accepted per ADR-0001.
- `GitState` ignores `GIT_DIR`/`GIT_WORK_TREE` *process* env that the old subprocess honored. This is **more** predictable (neutral-to-tightening), not a loosening; the command-string `GIT_DIR=` case remains a pre-documented exotic miss.

Pre-existing, out-of-scope (not introduced or widened here): the `cd <nonexistent> || true; git commit` unassessed-real-commit case and the cd-assumes-success behavior (#213/#224).

This satisfies the umbrella's mandatory security gate for the BLOCK guard.

cameronsjo:
## Security review — the gate caught a loosening, now fixed ✅

An adversarial Opus security-review pass (framed to hunt only for *loosening* — allow/nudge-where-block, the one severe failure mode for a BLOCK guard) reviewed the full diff, the guard, `GitState`, and its primitives. It did **not** pass clean the first time — it found one real loosening, which this PR now fixes and regression-locks.

### The finding — symlinked path defeated primary-checkout detection

`git rev-parse` discovers the repo by following symlinks **physically**; `find_git_root` walked **lexical** ancestors (`dir.pop()`). When a symlinked component makes a lexical ancestor own a `.git` before the physical repo is reached, they diverge:

> Layout: `repo/` primary; `outer/` a linked worktree of it; `outer/link` → `repo/src`. `git -C outer/link rev-parse --show-toplevel` → **`repo`** (primary). Lexical `find_git_root(outer/link)` → stops at **`outer`** (worktree, `.git` is a file) → `is_primary=false`. So `cd outer/link && git commit` physically commits into the primary but `assess_dir` judged the worktree → **ALLOW**, where git-based resolution **BLOCKED**.

### The fix

`GitState::resolve` now **canonicalizes `start` (resolving symlinks) before the lexical `find_git_root` walk**, so the walk climbs the physical ancestor chain git itself follows (`commit bae4cdf`). It also subsumes the #299 nonexistent-start check (`canonicalize` fails on a nonexistent path → `None`). A real-git regression test (`symlink_into_primary_resolves_to_primary_not_the_linking_worktree`) reproduces the exact exploit and asserts it resolves to the primary. All 130 enforce tests still pass; full workspace suite green.

### Everything else — cleared as no-loosening

| Vector | Finding |
|---|---|
| Same-repo comparison (`common_dir ==`) | Canonicalization makes the spawn-free path match git's `--path-format=absolute`; primary + its worktree share one canonical common dir (locked by test). |
| #299 nonexistent-start | Restores parity with git's `-C` failure; the un-fixed version would have *tightened* (false block), not loosened. No new TOCTOU. |
| `canonicalize_or` fallback | Same-repo paths resolve to the same physical `.git`, so the fallback fires for both or neither — no one-sided mismatch. |
| Snooze marker | Reader (GitState canonical) and git-based dismiss writer resolve to the same physical `.git` → same marker inode. No stale-honor / missed-active. |
| Submodules / nested / `.git` file-vs-dir / bare | Classify + resolve identically to git. |
| `GIT_DIR`/`GIT_WORK_TREE` ignored | Documented, neutral-to-tightening — not a loosening. |
| #271 immunity (zero spawns) | Removes a fail-open timeout window — tightening. |

With the symlink fix, the migration preserves or tightens the block decision on every reviewed vector. This satisfies the umbrella's mandatory security gate.

PR #335, opened 2026-07-20
fix(record-polish): print marker path on success (#257)

### LINKED ISSUE #257: feedback(record-polish): silent success — print the marker path so callers can verify the gate
`cadence-hooks cadence record-polish --scope full` exited 0 with no output when run in two worktrees, and a subsequent search could not locate the branch-scoped marker it claims to write (nothing polish-named under the repo's `.git/` or recently modified in the expected locations; only `~/.claude/metrics/polish_nudges.jsonl` turned up). Two problems:

1. **Silent success violates the CLI legibility convention** (payload to stdout, one unambiguous verdict line) — a caller can't distinguish "marker written" from "no-op'd quietly."
2. **The marker is unverifiable from outside** — the pre-PR gate (`nudge-polish-before-pr`) consumes it, but a session trying to confirm the gate is satisfied has no probe.

**Ask:** on success, print the marker's path (or a one-line `recorded: <repo>@<branch> scope=<scope>` verdict). A `--verify`/read subcommand would also do.

Observed 2026-07-07 on cadence-hooks 0.49.0 during a two-worktree pre-PR pass.

cameronsjo:
## What

`record-polish` now prints a one-line verdict on success: the marker path it wrote, plus the `(repo@branch, scope)` key. Previously the command exited 0 with no output, so a caller had no way to confirm where (or whether) the marker landed.

## How

- New pure formatter `record_verdict()` — unit-testable without stdout capture.
- The `write_marker` success branch prints the verdict to stdout; both failure paths keep their existing stderr lines verbatim.
- Fail-open semantics unchanged: exit 0 on every path, only the success path gains output.

No consumer parses this command's stdout (the pre-PR gate stats the marker file itself), so the new line is purely additive.

## Tests

- `record_verdict_names_path_repo_branch_and_scope` — asserts the verdict carries the marker path, the `repo@branch` key, and the scope.
- Existing marker-write coverage unchanged.

Closes #257

### DIFF
--- a/crates/cadence/src/record_polish.rs
+++ b/crates/cadence/src/record_polish.rs
@@ -25,6 +25,7 @@ use cadence_hooks_core::markers::{polish_marker, write_marker};
 use cadence_hooks_core::shell::git_command;
 use cadence_hooks_core::time::utc_timestamp;
 use serde_json::json;
+use std::path::Path;
 
 /// Resolve `repo_root`, `branch`, and `head_sha` from `dir`, letting explicit
 /// overrides bypass resolution so tests need no real repository.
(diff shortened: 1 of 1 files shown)


cameronsjo:
@coderabbitai review

cameronsjo:
Merging on the advisory-review fallback: automated review is rate-limited (~1 review/hour against this batch), and this repo treats that check as advisory. Verification behind this merge: CI green on both platforms; the full suite run independently against this branch matches the pristine-main baseline exactly (the two known environment-dependent doctor tests only); the one-function diff adds a stdout line on the success path only, verified to have no output-parsing consumers (the pre-PR gate stats the marker file).

PR #337, opened 2026-07-20
ci(release): cut workflow-artifact retention to 1 day

### LINKED ISSUE #301: ci(release): cut workflow-artifact retention on release.yml (release assets are the durable copy)
`release.yml` uploads the 5 platform binaries as workflow artifacts at the default 90-day retention on every `v*` tag. The GitHub Release assets are the durable copies — the workflow artifacts only need to survive the run itself.

**Proposed:** add `retention-days: 1` to the artifact upload step(s) in `release.yml`.

**Context:** account-wide Actions storage hit 100% of quota on 2026-07-10; the large repos were reclaimed and the heaviest workflow fixed (minute#375). This is the remaining hygiene item — ~stale binary artifacts accumulate a few hundred MB across release tags for no benefit.

cameronsjo:
## What

Sets `retention-days: 1` on the release workflow's artifact upload step. The five platform archives only bridge the `build` → `release` job handoff within a single run; the durable copies are the GitHub Release assets themselves.

## Why

Without an explicit value they inherit the 90-day default, keeping ~5 binaries per release on workflow-artifact storage long after they're dead weight. One day is GitHub's minimum and still leaves a full day's window to pull the archives when debugging a failed release run.

No other workflow in the repo uploads artifacts, and nothing downloads these outside the same run (the Homebrew dispatch parses `checksums.txt` inline, not the artifacts).

Closes #301

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -102,6 +102,7 @@ jobs:
         with:
           name: ${{ matrix.name }}
           path: ${{ env.ARCHIVE }}
+          retention-days: 1
 
   release:
     name: Create Release


cameronsjo:
Merging on the advisory-review fallback: automated review is rate-limited (~1 review/hour against this batch), and this repo treats that check as advisory. Verification behind this merge: CI green on both platforms; the diff is a single static YAML line (retention-days: 1) on the artifact-upload step, verified against the workflow's only consumer (the same-run release job) and the absence of artifact uploads in any sibling workflow.

PR #338, opened 2026-07-20
fix(doctor): liveness-aware --prune and actor-addressed --quiet warning

### LINKED ISSUE #305: doctor --prune --apply deletes version dirs live sessions are pinned to
## What happened

`cadence-hooks doctor --prune --apply` removed 95 orphaned plugin-cache version dirs (~216 MiB) — including the version dirs the **currently running Claude Code session was pinned to**. Every subsequent hook invocation in that live session failed (`Plugin directory does not exist: ~/.claude/plugins/cache/workbench/cadence/<sha> — run /plugin to reinstall`) until the user ran `/reload-plugins`. Observed 2026-07-12.

## Why

Claude Code resolves each plugin's hooks/skills/agents to a specific cache version dir at **session start** and holds those paths for the session's lifetime. A marketplace re-pin mid-day (new pushed commits) moves the manifest's current pin to a new dir, leaving the older generation on disk — which doctor's orphan check then classifies as prunable, because it compares only against the manifest pin. It has no notion of live sessions still executing from the older generation.

## Expected

"safe to prune — not the active pinned version" (doctor's own fix text) should mean safe — including for sessions currently running.

## Proposed fix

The binary already has a session registry (`cadence-hooks session status`, the same liveness data the canon plugin uses). Before deleting a version dir:

1. Check the registry for live sessions; if any exist, treat generations newer than the oldest live session's start as potentially referenced — or more simply, skip deletion entirely while any session is live unless `--force` is passed, with a message naming the live sessions.
2. The `.orphaned_at` marker observed in the output suggests a two-phase mark-then-delete already exists; if so, the delete phase is where the liveness check belongs (a grace period alone doesn't help — sessions can run for hours).

## Interaction protocol (also worth encoding in the fix text)

Because a prune can invalidate the *running* session's own hooks, the prune UX for an agent-driven run should be single-shot: confirm intent ("ready to run the prune — yes/no?"), execute exactly one tool call, and the very next output must instruct an immediate `/reload-plugins` **before any further tool calls**. Tonight's incident required exactly that recovery, discovered the hard way. Whatever fix lands, the `--apply` output should end with an explicit "if any Claude Code session is currently running, run /reload-plugins in it now" line.

### LINKED ISSUE #306: feedback(doctor): --quiet warning line names no actor, so it never gets actioned
## The line

Every session start injects (via the cadence plugin's SessionStart hook running `doctor --quiet`):

> cadence-hooks 0.55.0: 3 plugin warning(s) — run 'cadence-hooks doctor' for details (run 'brew upgrade cadence-hooks' (or 'cadence-hooks doctor' if already up-to-date))

## The problem

The line names no actor and asks for nothing, so nobody acts on it. It sat in every session's opening context for weeks, never actioned by any session — until a manual audit ran doctor and found 95 orphaned cache dirs (~216 MiB) plus 3 real warnings (a metrics stdin-parse cluster, a version_mismatch failopen, a missing pinned cache dir) that had been buried under orphan noise the whole time.

Semantically it reads as *status*, and its remedy is double-nested parentheses addressing two different audiences (the user for `brew upgrade`, ambiguously anyone for `doctor`).

## Proposed rewrite

Make the `--quiet` summary actor-explicit and imperative, e.g.:

> cadence-hooks: 3 warning(s). Claude: run `cadence-hooks doctor`, triage, and surface anything actionable to the user in one line.

Keep the `brew upgrade` hint only on the version-skew warning class where it's the actual remedy, not as a blanket parenthetical.

## Principle

A hook-injected nudge that names an actor and an action gets acted on; a status line gets read as wallpaper. (Same session's counter-example that works: the catalog blurb's "encourage the user to run a cadence:catalog preflight".)

Related: the prune-liveness issue filed alongside this one — actioning doctor's own prune advice mid-session broke the running session, so the doctor guidance this rewrite produces should also carry the single-shot + `/reload-plugins` protocol noted there.

cameronsjo:
Two `doctor` fixes, one commit each, plus a one-row docs entry.

## `--prune --apply` refuses while live sessions are pinned (#305)

The prune path computed orphans purely against the manifest pin and handed them straight to `remove_dir_all` — a live session still pinned to an "orphaned" version directory lost it out from under itself. `run_prune` now consults the session registry before deleting: any live session (including the invoking one — its own heartbeat counts) blocks the apply with a refusal that names the sessions, points at `/reload-plugins`, and names the escape hatch inline.

- Override: `CADENCE_DOCTOR_PRUNE_FORCE=1|true` forces through. The issue asked for a `--force` flag; the env override was chosen deliberately to keep the change confined to the doctor module — the refusal message makes it discoverable at exactly the moment it's needed. A flag can be layered on later if the ergonomics warrant.
- Known limitation, documented at the gate helper: the session registry is repo-scoped while the plugin cache is global, so sessions in *other* repos aren't seen. The gate is under-protective across repos, never over-destructive — and it reliably catches the incident case (the session running the prune).
- The `--apply` success output now reminds the operator to run `/reload-plugins` in any live session.

## `--quiet` warning names an actor and action (#306)

The one-line summary a quiet doctor run emits was actor-free and blanket-appended an upgrade hint regardless of warning class. It now addresses the actor with a concrete next step, and the upgrade hint appears only when a version-skew warning is actually present.

## Tests

Eight new inline tests: five on the liveness gate (force-through, missing dir, empty dir, live-session blocks-and-names, stale-only proceeds) and three on the quiet summary (actor+action present, hint omitted without skew, hint present with skew).

Closes #305
Closes #306

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -109,6 +109,7 @@ kept unprefixed because it's a cross-tool convention.
 | `CADENCE_METRICS_DEBUG` | `log-subagent` | Set to `1` to append a `_keys` array of the raw payload's top-level keys to subagent records — surfaces schema additions across Claude Code releases |
 | `CADENCE_METRICS_STALE_DAYS` | `warn-stale` | Days of metrics-write silence before the SessionStart alarm fires and `doctor` reports staleness (default 4); zero or unparseable falls back to the default |
 | `CADENCE_SESSION_STALE_MINUTES` | `session` hooks | Minutes of heartbeat silence before a session is presumed dead (default 10) |
+| `CADENCE_DOCTOR_PRUNE_FORCE` | `doctor --prune` | Set to `1` or `true` to bypass the live-session gate and let `doctor --prune --apply` delete orphaned plugin-cache version dirs even while peer sessions are running. The gate's refusal message names this override; otherwise run `/reload-plugins` in the live session first to release the retired dirs |
 | `GH_AUTOCLOSE_WAIT_SECONDS` | `verify-pr-autoclose` | Seconds to wait after `gh pr merge` before checking for straggler issues (default 10) |
 | `OBSIDIAN_VAULT` | `trash-guard`, `warn-overshare` | Absolute path to Obsidian vault — the trash guard scopes `rm` blocking to it, and the overshare audit treats it as the safe destination for personal context |
 
--- a/src/doctor.rs
+++ b/src/doctor.rs
@@ -17,6 +17,9 @@ use std::path::{Path, PathBuf};
 use std::time::{Duration, SystemTime};
 
 use crate::registry;
+// The session crate's `registry` (peer-session liveness) — aliased because the
+// bare name is already taken by the binary's hook-catalog `registry` above.
+use cadence_hooks_session::registry as session_registry;
 
 /// Severity of a finding: errors block (shell bugs), warnings advise (version skew).
 #[derive(Debug, PartialEq)]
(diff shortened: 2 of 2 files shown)


cameronsjo:
Merging on the advisory-review fallback: automated review is rate-limited (~1 review/hour against this batch), and this repo treats that check as advisory. Verification behind this merge: CI green on both platforms; the full suite run independently against this branch matches the pristine-main baseline exactly (the two known environment-dependent doctor tests only); the prune change moves in the conservative direction only (block-by-default with a named override) with eight new tests on the gate and summary builders.

PR #339, opened 2026-07-20
fix(core): tolerate JSONC comments and trailing commas in settings.json reads (#246)

### LINKED ISSUE #246: repo_env_flag: tolerate JSONC (comments / trailing commas) in settings.json reads (follow-up to #239 F10)
## Follow-up from #239 F10

#239 made `repo_env_flag` strip a leading UTF-8 BOM before `serde_json::from_str`, fixing the common editor-BOM case where a declared `CADENCE_ALLOW_MAIN` was silently dropped and a by-design-main repo false-blocked.

The narrower gap remains: a repo's `.claude/settings.json` containing **JSONC** (`//` or `/* */` comments, or a trailing comma) still fails strict `serde_json` parsing → the declared exemption is discarded → the repo false-blocks. By the friction-budget principle (a fail-open guard spends its budget on false-*allows*, never false-*blocks* — cadence-forge:developing-guards), this is worth closing.

Deferred from #239 because it needs a dependency decision (a lenient parser — `json5`, `jsonc-parser`, or a hand-rolled comment/trailing-comma stripper) rather than the one-line BOM strip. Whatever lands must stay within the existing bounded/hardened read (`read_untrusted_config`: regular-file check, 1 MiB cap) and remain fail-open (a parse failure = "not declared", never a block or panic).

Scope: `crates/core/src/config.rs` `repo_env_flag` (and any other reader that would benefit — the #164 guard-family settings generalization is the eventual home). Low urgency; hand-authored JSONC in a Claude settings file is uncommon, but silent when it bites.

Related: #239 (F10 BOM strip), #232 (repo-scoped ALLOW_MAIN), #164 (settings-resolution generalization).

cameronsjo:
## What

Settings reads now tolerate JSONC — `//` and `/* */` comments and trailing commas — instead of silently discarding the whole file on a strict-parse failure. A declared flag in a commented settings file previously vanished, false-blocking repos that had legitimately opted in.

## How

- New module-private `parse_jsonc`: BOM strip → string-aware comment/trailing-comma stripping → strict `serde_json` parse of the cleaned text. Two clear passes, each tracking string-literal and escape state, so `//` inside a URL value or a `,` inside a string survive verbatim.
- `repo_env_flag` and `load_cadence_section` both route through it (the latter gains BOM tolerance it never had). Fail-open contracts are unchanged: genuinely malformed input still yields `None` / `T::default()` — a stripper bug can only degrade to today's behavior, never fabricate a truthy value.
- No new dependency — the needed JSONC surface is tiny and the repo's bias is minimal deps.
- Deliberately untouched: the settings **write** path (`configure.rs`) — tolerating comments on a read-modify-write path would silently strip them on rewrite; that consolidation needs its own design.

## Tests

19 new inline cases: tolerance for line/block comments, trailing commas, BOM+JSONC combined; footgun fixtures proving string interiors are never mangled (`http://` values, commas/braces in strings, escaped quotes); fail-open regressions (unterminated/nested comments, comment-only files, malformed JSON still fall through).

Closes #246

### DIFF
--- a/crates/core/src/config.rs
+++ b/crates/core/src/config.rs
@@ -230,14 +230,13 @@ pub fn repo_env_flag(repo_root: &Path, key: &str) -> Option<String> {
         let Some(content) = read_untrusted_config(&claude_dir.join(name)) else {
             continue;
         };
-        // Tolerate a leading UTF-8 BOM. Editors and hand-authoring often prepend
-        // one; serde_json rejects it as an unexpected character at column 1,
-        // which would silently drop a *declared* CADENCE_ALLOW_MAIN and
-        // false-block a by-design-main repo (cadence-hooks#239 F10). We strip
-        // only the BOM — JSONC comments/trailing commas stay a strict-parse
-        // reject (no lenient-parser dependency here).
-        let content = content.strip_prefix('\u{feff}').unwrap_or(&content);
-        let Ok(json) = serde_json::from_str::<serde_json::Value>(content) else {
+        // Tolerate a leading UTF-8 BOM and JSONC (comments, trailing commas).
+        // Editors and hand-authoring often prepend a BOM or leave `//` comments;
+        // strict serde_json rejects both, which would silently drop a *declared*
+        // CADENCE_ALLOW_MAIN and false-block a by-design-main repo
+        // (cadence-hooks#239 F10, #246). parse_jsonc strips them before parsing;
+        // still-malformed input yields None and falls through.
+        let Some(json) = parse_jsonc(&content) else {
             continue;
         };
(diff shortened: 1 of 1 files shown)


cameronsjo:
@coderabbitai review

cameronsjo:
Merging on the advisory-review fallback: automated review is rate-limited (~1 review/hour against this batch), and this repo treats that check as advisory. Verification behind this merge: CI green on both platforms; the full test suite run independently against this branch and against pristine main produced identical failure sets (the two known environment-dependent doctor tests only); the change was adversarially verified against the issue's claims at plan stage, and the parser had a dedicated security review pass (string-aware stripping, fail-open bounds).

PR #340, opened 2026-07-20
fix(secret-leaks): scope the echo/printf secret nudge to expanded secret-shaped vars (#332 #333 #334 #321)

### LINKED ISSUE #321: feedback(cadence-hooks): secret-env print nudge fires on benign status echoes (~7×/session)
## Problem

The "⚠️ Command may print a secret environment variable. Run programs that use env vars directly instead." PreToolUse nudge fired ~6 times in one session (2026-07-18, dotfiles slim-down) on benign commands — e.g. `echo "DIFF_RC=$?"` status echoes, `stat`-arithmetic token counts, a `grep` over a transcript. None referenced a secret-bearing variable.

## Lane

Fine-tune (cadence-hooks) — the guard's intent is right; its trigger reads as "any command that interpolates any variable into output", which is most shell one-liners.

## Reproduction

Payload shape that fired: `{"tool_name":"Bash","tool_input":{"command":"chezmoi diff ~/.claude/CLAUDE.md; echo \"DIFF_RC=$?\"; ..."}}` — the only expansions are `$?` and locally-assigned loop variables.

## Expected vs actual

Expected: the nudge fires when a command would print a variable plausibly holding a secret (name matching KEY/TOKEN/SECRET/PASSWORD patterns, or values sourced from env). Actual: it fires on `$?`, `$((...))` arithmetic, and locally-assigned scratch vars, so the signal drowns.

## Suggested Fix

Maintainer's call — likely tightening the variable-name heuristic (skip `$?`, `$#`, `$@`, arithmetic expansion, and same-command locally-assigned vars).

### LINKED ISSUE #332: feedback(secret-env warning): high false-positive rate on benign $?/PATH-prefix/exit-probe commands
## Friction

The "⚠️ Command may print a secret environment variable. Run programs that use env vars directly instead." PreToolUse warning fired repeatedly on benign commands through a long working session (2026-07-19, session umber-wheel): multi-command chains containing `git commit -m` heredoc-style messages, `echo "commit: $?"` exit-code probes, and `PATH="$HOME/.cargo/bin:$PATH" cargo …` prefixes. None printed secret values; the trigger appears to be shell-variable expansion patterns (`$?`, `$HOME`, assignments like `S=…; …`) anywhere in the command string.

At this false-positive rate the warning trains the reader to ignore it, which spends the credibility the real catches need.

## Fine-tune ideas

- Exclude the boring expansions: `$?`, `$#`, `$HOME`, `$PWD`, `$PATH`-prefix assignments, and single-variable `echo "<label>: $?"` probes.
- Only fire when the expanded variable name matches secret-shaped patterns (KEY/TOKEN/SECRET/PASSWORD/CREDENTIAL) or when the command is an env-dumper (`env`, `printenv`, `set`) — the sibling prevent-secret-leaks checks already model this well.
- Consider rate-limiting per session for identical command shapes.

Filed from the outro dogfood sweep.

### LINKED ISSUE #333: feedback(env-shape-guard): 4 false positives on benign git/gh commands in one session
## Summary

The may-print-secret-env-var heuristic fired 4 times in one Claude Code session (2026-07-19), all false positives — none of the flagged commands involved secrets.

## Observed false positives

1. A `git commit` invocation whose commit message contained a GitHub noreply email address (the `NNNN+user@users.noreply.github.com` pattern) — the pattern apparently resembles an env assignment or credential.
2. `VAR=$(gh pr view N --json body ...)` — capturing PR-body text into a shell variable, followed by file writes. Flagged twice in the same session.
3. A heredoc-fed `git commit` where the message body simply contained words like "email" / "config".

Each firing was PreToolUse advisory context, not a hard block, so the cost is noise rather than a stopped workflow — but 4 false positives in one session erodes trust in the signal.

## Suggested tunings

- Allowlist the `users.noreply.github.com` literal outright.
- Skip the check when the command is a `git commit -m` / `--body-file` invocation and the suspicious text sits inside a quoted commit message or heredoc body, rather than an interpolated shell variable.
- Alternatively, require the pattern to actually look like `NAME=value` at command position rather than matching substrings anywhere in the command line.

Happy to reproduce sanitized sample commands if that's useful for a regression test.

### LINKED ISSUE #334: feedback(env-shape-heuristic): markdown heredoc append flagged as potential secret dump
**What happened:** A PreToolUse advisory warned "Command may print a secret environment variable" on a plain markdown append:

```bash
command cat >> "/path/to/notes/Log.md" <<'EOF'
## [2026-07-19] Update | ...
- Updated: ...
EOF
echo "LOG_APPENDED $?"
```

**Expected:** No warning — the heredoc body is prose/markdown (headings, list items, wiki-links), contains no env-var-shaped tokens, and the only echo prints a literal sentinel plus `$?`.

**Suspected cause:** The env-shape heuristic appears to trigger on the heredoc + echo combination without checking the payload for env-var density (e.g. `FOO=bar` lines, `${VAR}` expansions, known-sensitive prefixes).

**Suggested fine-tune:** Require at least one env-var-shaped signal inside the heredoc body (assignment-shaped lines, brace expansions, or known-sensitive variable prefixes) before warning; a body that is markdown-shaped (leading `#`/`-` lines) could weigh against.

Severity: advisory-only, no block — filed as a fine-tune, not a bug.
If this fired in error, ignore — this IS the feedback.

cameronsjo:
## Root cause — one branch, four issues

The "may print a secret environment variable" nudge fired when an `echo`/`printf` appeared as the command word of *any* segment **and** one of the six secret keywords appeared as a bare substring *anywhere in the whole command*. The two conditions were decoupled: a keyword inside a commit message, a heredoc body, a filename, or "author" satisfied the second while the echo printed only `$?`. That combination fired ~7×/session on routine chains (`git commit -m "fix(secret): …"; echo "commit: $?"`).

## Fix

The predicate is now segment-scoped: the nudge fires only when the `echo`/`printf` segment's own arguments expand a secret-shaped variable (`$NAME`/`${NAME}` whose name contains one of the same six keywords). The keyword set and the nudge message are unchanged.

## What stays caught

- `echo $API_KEY | curl -d @- …` and `printf '%s' "$GITHUB_TOKEN" > file` — the canonical exfil shapes, pinned by new regression tests.
- The six pre-existing warned tests pass unmodified.
- The env-dump branch (`env`, `printenv`, `export -p`) and the `.env` file-read BLOCK branch are untouched.

Anything newly quiet was a pre-existing gap dressed differently (e.g. secrets in non-keyword-named vars never fired under the old predicate either) — no real detection surface is lost, and the check remains advisory.

## Tests

Eight new: six quiet cases reproducing the exact reported shapes (commit-message keyword, `DIFF_RC=$?` chains, PATH prefix, literal "token count", command-substitution capture, heredoc append) and two still-caught exfil shapes.

Closes #332
Closes #333
Closes #334
Closes #321

### DIFF
--- a/crates/cadence/src/prevent_secret_leaks.rs
+++ b/crates/cadence/src/prevent_secret_leaks.rs
@@ -9,11 +9,20 @@
 
 use crate::secret_patterns::{
     command_may_reference_secret, envrc_carveout_allows, is_ambiguous, is_blocked,
-    is_dangerous_secret_token, is_safe_template,
+    is_dangerous_secret_token, is_safe_template, is_secret_shaped_var_name,
 };
 use cadence_hooks_core::shell::{command_segments, split_segments, tokenize};
 use cadence_hooks_core::{Check, CheckResult, HookInput};
+use regex::Regex;
 use std::path::Path;
+use std::sync::LazyLock;
+
+/// Captures the NAME of a shell variable expansion (`$VAR`, `${VAR`) — the
+/// leading `$`, an optional `{`, then a valid identifier. Same identifier
+/// family as `validate_env_vars`'s access pattern. Used to judge whether an
+/// echo/printf argument expands a secret-shaped variable.
+static VAR_EXPANSION_PATTERN: LazyLock<Regex> =
+    LazyLock::new(|| Regex::new(r"\$\{?([A-Za-z_][A-Za-z0-9_]*)").expect("var pattern compiles"));
 
 /// Commands that only touch file metadata — they never emit file contents,
 /// so a `.env` operand is safe. `cp`/`mv`/`ln`/`tar` are deliberately NOT
@@ -228,6 +237,33 @@ fn original_case_token_at(
     Some((original.to_string(), end))
 }
 
+/// Does an `echo`/`printf` segment expand a secret-shaped variable?
+///
+/// Segment-scoped on purpose: the keyword must live in a variable EXPANDED by
+/// an `echo`/`printf` in the SAME segment, not anywhere in the whole command.
+/// That decoupling is the #332/#333/#334/#321 fix — the prior whole-command
+/// keyword substring check fired whenever any keyword appeared alongside an
+/// echo/printf elsewhere in the chain. Mirrors [`is_executed_command`]'s
+/// per-segment, group-punctuation-trimming command-word detection so a benign
+/// arg or path containing "echo"/"printf" doesn't over-fire.
+fn echo_or_printf_leaks_secret_var(lower: &str) -> bool {
+    for segment in split_segments(lower) {
+        let segment = segment.trim_start_matches(['(', '{', ' ', '\t']);
+        match segment.split_whitespace().next() {
+            Some("echo") | Some("printf") => {}
+            _ => continue,
+        }
+        if VAR_EXPANSION_PATTERN
+            .captures_iter(segment)
+            .filter_map(|c| c.get(1))
+            .any(|name| is_secret_shaped_var_name(name.as_str()))
+        {
+            return true;
+        }
+    }
+    false
+}
+
 /// Check if a bash command would dump secrets to stdout.
 ///
 /// `cwd` is the tool call's working directory, used only to resolve a relative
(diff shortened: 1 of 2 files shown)


cameronsjo:
Merging on a two-lens review panel (in lieu of the rate-limited automated reviewer).

Security lens (false-negative hunt on the narrowed nudge): CLEAN — echo/printf position detection is byte-identical between old and new; only the keyword test narrowed (whole-command substring → a secret-shaped variable the echo/printf segment actually expands). All canonical exfil shapes still nudge (`echo $API_KEY | curl`, `printf '%s' "$GITHUB_TOKEN"`, `echo "${DB_PASSWORD}"`), the .env-file BLOCK and env-dump branches are untouched, and the only shapes newly quiet were coincidental keyword-elsewhere catches the old predicate never reliably covered.

Correctness lens: PASS — the SECRET_VAR_NAME_KEYWORDS hoist is byte-identical to the prior inline list, the regex handles $NAME and ${NAME}, tests cover both the fixed false-positive matrix and a still-warned matrix. Remaining nits (single-quote/env-assignment blind spots) are pre-existing parity, over-nudge friction, never a bypass. Suite green (192 cadence), CI green both platforms.

PR #341, opened 2026-07-20
fix(guard-rm): allow file-scoped artifact/temp deletes and correct the bypass hint (#322 #316 #311)

### LINKED ISSUE #311: feedback(guard-rm): CADENCE_BYPASS=1 hint is unusable as a command env prefix — the hook never sees it
## Problem

`guard-rm`'s block message offers `CADENCE_BYPASS=1 — bypass all cadence enforcement for one session` as a remediation. Retrying the blocked command with that as an env prefix (`CADENCE_BYPASS=1 rm -rf <repo-dirs>`) blocks identically — the hint is unusable from an agent-driven shell invocation, which is exactly where the guard fires.

## Diagnosis

2026-07-14, during a sanctioned cleanup of retired plugin checkouts (user-approved, evidence-gated):

1. `rm -rf <nine retired repo checkouts>` → blocked by guard-rm (correct — the dirs contain `.git`).
2. Retry as `CADENCE_BYPASS=1 rm -rf <same dirs>` → blocked with the identical message.

## Root Cause

PreToolUse hooks run in their own process environment and receive the command as a string. An env assignment prefixed inside that string is shell syntax the *command's* subshell would apply — the hook process never sees it, and the guard doesn't parse the command string for a leading bypass assignment. The hint implicitly assumes the variable is exported in the environment Claude Code itself runs under (an interactive `export`, or a launch-time env), which an in-session retry can't achieve.

## Suggested Fix

Likely approaches (any or all):

- Recognize a leading `CADENCE_BYPASS=1 ` env-prefix in the inspected command string as the documented one-shot bypass (making the hint true as written), or
- Reword the hint to state where the variable must live ("export it in the environment Claude Code runs under, or use `CADENCE_DISABLE=guard-rm` in settings") so the remediation is actionable from the surface that hit the block, and/or
- Advertise the move-based alternative in the hint: `mv` to `~/.Trash` is permitted by the guard, recoverable, and was the successful path in this case — arguably the better first remediation for "delete a repo" intent.

Confidence: root cause confirmed empirically (two identical blocks, one with the prefix); fix options are suggestions, maintainer's call on which.

## Labels

feedback

### LINKED ISSUE #316: feedback(cadence-hooks): guard-rm blocks atomic-write temp cleanup (~/.claude.json.tmp) despite allow-temp branch
## Problem

`guard-rm` (0.59.0) hard-blocks `rm` of an atomic-write temp file that lives directly in `$HOME`, even though the guard's own description promises "allow temp/managed". The top-level-home-dir rule fires before the temp-allow can exempt the path.

Encountered doing an atomic edit of `~/.claude.json` (jq del → temp → validate → `mv`), with a failure-branch cleanup:

```bash
jq 'del(.mcpServers.kanban)' ~/.claude.json > ~/.claude.json.tmp && \
  jq -e . ~/.claude.json.tmp >/dev/null && \
  mv ~/.claude.json.tmp ~/.claude.json || { rm -f ~/.claude.json.tmp; }
```

guard-rm blocked the **entire** command (so the jq/mv that should have run never did) because `rm ~/.claude.json.tmp` "deletes a top-level entry in your home directory."

## Reproduction

```
cadence-hooks --version   # 0.59.0
printf '{"tool_name":"Bash","tool_input":{"command":"rm -f ~/.claude.json.tmp"},"cwd":"%s"}' "$HOME" > /tmp/pf.json
cadence-hooks try guardrails guard-rm --payload /tmp/pf.json --show-payload
```

Outcome:

```
Outcome:  BLOCK (exit 2)
Stderr:   🚫 guard-rm: this deletes a top-level entry in your home directory, which is almost never intended.
```

## Expected vs actual

- **Expected:** allow — `~/.claude.json.tmp` is a `.tmp` scratch sibling of an atomic-write target, the canonical temp-then-`mv` pattern the guard's "allow temp/managed" branch is meant to cover.
- **Actual:** blocked as a top-level home-dir deletion; the whole compound command aborted.

## Suggestion (maintainer's call)

Consider exempting an obvious scratch sibling from the top-level-home block — an `rm` whose target ends in `.tmp`/`.bak`/`.swp`, or matches `<existing-file>.tmp`. The workaround (drop the `rm`, leave the temp file behind) is cheap but forces awkward command shapes and leaves litter.

Minor friction, cheap workaround — not urgent.

_Filed via cadence:feedback (guard-rm false-positive)._

### LINKED ISSUE #322: feedback(cadence-hooks): guard-rm blocks artifact-glob deletes on mere presence of .git in cwd
## Problem

`guard-rm` blocks any `rm` executed in a directory containing a `.git`, regardless of what the `rm` actually targets. A build-and-repack chain whose only delete is an npm-pack artifact glob (`homebridge-dreo-*.tgz`) is hard-blocked as "deletes a git repository."

## Environment

- `cadence-hooks 0.59.0`
- Hook: `guardrails guard-rm`
- Owner: `cameronsjo` (own repo), no relevant `CADENCE_*` overrides in play

## Reproduction

Payload (rebuilt from the real triggering command):

```json
{"tool_name":"Bash","tool_input":{"command":"npm run build 2>&1 | tail -5 && rm -f homebridge-dreo-*.tgz && npm pack"},"cwd":"~/Projects/homebridge/dreo"}
```

Run:

```
cadence-hooks try guardrails guard-rm --payload rm-payload.json --show-payload
```

Outcome: **BLOCK (exit 2)**

```
🚫 guard-rm: this deletes a git repository (a `.git` lives here), which is almost never intended.
```

The `rm -f homebridge-dreo-*.tgz` targets only npm-pack tarballs in the repo root — it can never match `.git` or the repo directory itself.

## Expected vs actual

- **Expected:** allow — the `rm` target is a `*.tgz` artifact glob, not the repository or `.git`.
- **Actual:** hard block, on the mere presence of a `.git` in `cwd`.

## Suggested seam

The block appears keyed on "a `.git` lives in `cwd`" rather than on the `rm` target set. Inspecting the actual target paths/globs — blocking only when the resolved targets could include `.git` or the repo root — would let artifact-cleanup deletes through.

cameronsjo:
Three guard-rm fixes, one commit each.

## Truthful bypass hint (#311)

The block message implied `CADENCE_BYPASS=1` works as a command-line prefix. It can't: the hook reads its **own** environment, never the command string, so a prefix applies only to the command's subshell — invisible to the guard. The message now says the variable must live in Claude Code's environment (settings `env` block or a pre-launch export), and offers the recoverable alternative `mv <target> ~/.Trash` (guard-rm doesn't fire on `mv`). Message-only; the parser is shared by every guard and is correct.

## Transient-suffix scratch deletes in home (#316)

`rm -f ~/.claude.json.tmp` (the atomic-write cleanup idiom) was blocked because first-level home entries classified purely by location. A new `Scratch` class allows final segments ending in `.tmp`/`.swp`/`.swo` — unambiguous scratch conventions. `.bak`/`.orig`/`.old` are deliberately excluded (often intentional backups). `~/.zshrc`, `~/Documents`, `~/Documents.bak`, `~/.claude` all still block.

## File-scoped globs classify by their directory (#322)

`rm -f something-*.tgz` at a repo root was blocked as "delete the git repo": the glob's literal prefix reduced to the cwd, which classified as the repo directory itself. Globs whose final segment is a real file filter (contains a metachar, doesn't start with `.`, isn't a bare `*`/`?`) now classify by their *directory* with the outcome split by recursion: non-recursive → allow (can only unlink matching files — never `.git`, dot-entries, or directories), recursive (`rm -rf project-*`) → ask, human-gated. Bare `rm -rf *`, dot-globs (`.*`), vault globs, and `/*` all still block — pinned by tests.

## Tests

Fifteen new inline cases: the reported false positives quiet, and a still-blocked matrix across repo-wipe, dot-glob, vault, home, and root shapes.

Closes #322
Closes #316
Closes #311

### DIFF
--- a/crates/guardrails/src/guard_rm.rs
+++ b/crates/guardrails/src/guard_rm.rs
@@ -89,6 +89,10 @@ enum TargetClass {
     /// is NOT here — deleting it whole is not the transient-scratch case this
     /// ALLOW exists for, so it falls through to the block/ask rules.
     ClaudeManaged,
+    /// A transient scratch/editor-swap file directly under home whose name ends
+    /// in a recognized ephemeral suffix (`.tmp`/`.swp`/`.swo`). ALLOW — deleting
+    /// these is routine cleanup, not the destructive home-child case.
+    Scratch,
     /// The filesystem root `/`. BLOCK.
     Root,
     /// The user's home directory itself. BLOCK.
@@ -110,7 +114,7 @@ enum TargetClass {
 impl TargetClass {
     fn outcome(self) -> Outcome {
         match self {
-            TargetClass::Temp | TargetClass::ClaudeManaged => Outcome::Allow,
+            TargetClass::Temp | TargetClass::ClaudeManaged | TargetClass::Scratch => Outcome::Allow,
             TargetClass::Root
             | TargetClass::Home
             | TargetClass::HomeChild
@@ -384,6 +388,15 @@ fn has_parent_segment(path: &str) -> bool {
     path.split('/').any(|seg| seg == "..")
 }
 
+/// True when `norm`'s final path segment ends in a recognized transient suffix
+/// (`.tmp`, `.swp`, `.swo`) — an editor swap file or a temp-write artifact.
+/// `.bak`/`.orig`/`.old` are deliberately excluded: those can be intentional
+/// backups a user means to keep, so they stay BLOCK under home.
+fn is_transient_scratch(norm: &str) -> bool {
+    let last = norm.rsplit('/').next().unwrap_or(norm);
+    last.ends_with(".tmp") || last.ends_with(".swp") || last.ends_with(".swo")
+}
+
 /// Classify a resolved target path into guard-rm's [`TargetClass`].
 ///
 /// Delegates the shared path facts to [`pathclass::classify`] — temp,
@@ -425,6 +438,11 @@ fn classify_path(path: &str, ctx: &RmContext, is_git_root: &dyn Fn(&str) -> bool
         return TargetClass::Home;
     }
     if shared == PathClass::HomeChild {
+        // A transient scratch/swap file under home (`~/.claude.json.tmp`,
+        // `~/.foo.swp`) is routine cleanup, not the destructive home-child case.
+        if is_transient_scratch(&norm) {
+            return TargetClass::Scratch;
+        }
         return TargetClass::HomeChild;
     }
     // Vault is guard-rm-local (deferred from pathclass v1); checked ahead of the
@@ -679,6 +697,23 @@ mod tests {
     fn deep_home_path_is_not_child() {
         // Only first-level home entries block; deeper paths fall to the middle.
         assert_eq!(judge("rm -rf ~/Documents/notes", "/home"), Outcome::Ask);
+    }
+
+    // --- ALLOW: transient-suffix scratch files directly under home (#316) ---
+
+    #[test]
+    fn transient_scratch_under_home_allows() {
+        // Editor swaps and temp-write artifacts under home are routine cleanup.
+        assert_eq!(judge("rm -f ~/.claude.json.tmp", "/home"), Outcome::Allow);
+        assert_eq!(judge("rm -f ~/.foo.swp", "/home"), Outcome::Allow);
+    }
+
+    #[test]
+    fn non_transient_home_child_still_blocks() {
+        assert_eq!(judge("rm -rf ~/.zshrc", "/home"), Outcome::Block);
+        assert_eq!(judge("rm -rf ~/Documents", "/home"), Outcome::Block);
+        // `.bak` is a deliberate exclusion — it can be an intentional backup.
+        assert_eq!(judge("rm -rf ~/Documents.bak", "/home"), Outcome::Block);
     }
 
     // --- BLOCK: vault ---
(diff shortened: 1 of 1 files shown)


cameronsjo:
Merging after a two-lens review panel plus a hardening pass (in lieu of the rate-limited automated reviewer).

Security lens (adversarial, destructive-rm hunt): CLEAN — both softenings correctly bounded. Every destructive shape traced to Block/Ask: bare `rm -rf *`/`.*` still Block, home/vault globs Block, `../`/`$VAR`/backtick operands Ask (ahead of glob detection), `.bak`/`.orig` suffixes Block, and the `[.]env` character-class dotfile dodge is a shell no-op. Non-recursive rm structurally can't delete directories, so the file-glob Allow path removes only files.

Correctness lens: PASS — new enum variants thread through exhaustive match arms, tests assert the full still-blocked matrix. The lens flagged one recursion-detection gap: `rm_is_recursive` matched `--recursive` exactly, missing GNU's unambiguous abbreviations (`--rec`, `--r`) — a `--recu` file-glob sweep could ride Ask→Allow. Hardened in 092df3e (any `--<prefix-of-recursive>` now counts as recursive), with a security regression test proving `rm --recu project-*` in a git repo → Ask, not Allow. Guard suite 64/64, CI green both platforms.

Accepted residual (documented follow-up): a non-recursive file-scoped glob in a repo softens for any extension, so `rm -f *.pem` can delete untracked secrets git won't recover — within the fail-open discipline-guard charter.

PR #342, opened 2026-07-20
feat(guard-gh-write): unblock read-only and curated review-thread graphql, keep arbitrary mutations closed (#262 #263 #300 #317)

### LINKED ISSUE #262: feedback(cadence-hooks): guard-gh-write fails closed on 'gh api graphql' (-R not accepted, cwd->owner fallback misses live)
## Problem

`guardrails guard-gh-write` blocks `gh api graphql` writes to an **allowed** owner, and the remediation it prints (`add -R owner/repo`) cannot be applied — `gh api graphql` has no `-R`/`--repo` flag (nor does the guard honor `GH_REPO=` or `-F repo=` as a substitute). REST calls to the same owner are correctly allowed because the owner is parseable from the `repos/OWNER/REPO/...` endpoint path. The net effect: any GraphQL mutation on an owned repo (e.g. `resolveReviewThread` to resolve a PR review thread) is un-runnable through the guard, with no actionable escape short of disabling the hook.

## Diagnosis

Version: `cadence-hooks 0.52.0`. Owner is in `CADENCE_ALLOWED_OWNERS`; `gh api user --jq .login` confirms the account is a collaborator on the target — so the block is a false positive, not a real ownership violation.

Observed behavior (live PreToolUse hook), cwd = a git checkout whose `origin` remote owner is in `CADENCE_ALLOWED_OWNERS`:

- `gh api graphql -f query='{viewer{login}}'` → **BLOCKED**: `Cannot determine target repo for gh write operation`, with the correct `Directory:` (the repo checkout) shown in the message.
- `gh api graphql -R OWNER/REPO ...` → guard **allows**, but `gh` then errors `unknown shorthand flag: 'R'` — the suggested fix is not a valid `gh api graphql` flag.
- `GH_REPO=OWNER/REPO gh api graphql ...` and `-F repo=OWNER/REPO ...` → still **BLOCKED**.
- `gh api repos/OWNER/REPO/...` (REST, owner in path) → **allowed**.

The tell — **not reproducible via `cadence-hooks try`**:

```
cadence-hooks try guardrails guard-gh-write --payload <same command + same cwd>
→ Outcome: ALLOW (exit 0)
```

`try` (unsandboxed) resolves cwd → `git remote` → owner → allowed. The **live** hook, with the identical command and cwd, cannot, and fails closed. `git remote -v` in that cwd resolves cleanly to the allowed owner in both contexts.

## Root Cause

*(hypothesis — reporting the seam, not prescribing the fix)* The guard derives the target owner for a `gh api graphql` call (which carries no owner in the command) from the cwd `git remote`. That fallback resolves under `cadence-hooks try` but not in the live PreToolUse environment, so an allowed-owner GraphQL write fails closed. The two working signals — `-R` and the REST `repos/OWNER/REPO` path — are the only owner sources the live guard accepts, and `gh api graphql` can supply neither.

## Suggested Fix

Needs-investigation, not a confirmed patch. Directions worth considering:
- Make the cwd → owner resolution work (or degrade to allow-for-allowed-owner) in the live hook environment, matching `cadence-hooks try`.
- Recognize an owner from `-f owner=...` / query text, or accept an env signal (`GH_REPO`), for `gh api graphql`.
- At minimum, correct the remediation text: `-R` is invalid for `gh api graphql`, so the printed "add `-R owner/repo`" hint is un-actionable for the exact command class that triggers this block.

## Labels

cadence-hooks, bug

### LINKED ISSUE #263: feedback(guard-gh-write): read-only GraphQL (gh api graphql query) is blocked as an unverifiable write
**What happened:** a pure schema-introspection read was hard-blocked:

```
gh api graphql -f query='query { __type(name: "Mutation") { fields { name } } }'
🚫 git-guardrails: gh api write to an unverifiable target — ownership can't be checked
```

**Why it fired:** `gh api graphql` is a POST to a non-`repos/...` path, so the guard classifies it as a write to an unverifiable target. Fail-closed is the right default — GraphQL mutations absolutely belong behind the guard.

**Why it matters now:** GitHub Projects v2 is becoming the live work queue (claude-configurations ADR-0035, cadence#322), and Projects reads beyond `gh project`'s porcelain (field values in one query, item filtering, schema probes) are GraphQL-only. Every such read currently blocks.

**Proposed carve-out:** allow `gh api graphql` when the operation is verifiably query-only — the `query=` payload's first non-whitespace token is `query` or `{` (and contains no `mutation` token). GraphQL requires mutations to be declared with the `mutation` keyword, so this classification is sound at the syntax level. Everything else (any `mutation`, undecidable payloads, `-f`-composed variables that obscure the operation) keeps blocking.

**Workaround meanwhile:** `gh project item-list/field-list/view` porcelain covers the common reads; this is about the residue.

If this fired in error: it did not — the guard behaved as designed; this is a feature request for a sound read-only carve-out.

### LINKED ISSUE #300: feedback(guard-gh-write): resolveReviewThread is GraphQL-only — the suggested repos/ fix is unsatisfiable
## What happened

`guard-gh-write` blocked a `gh api graphql` call carrying a `resolveReviewThread` mutation:

```
🚫 git-guardrails: gh api write to an unverifiable target — ownership can't be checked
   Fix: use `gh api repos/<owner>/<repo>/…` so ownership is checkable, or ask the user
```

The block is correct under the guard's current rule (GraphQL bodies aren't parsed, so the target repo is unverifiable). The friction: **thread resolution has no REST equivalent** — `resolveReviewThread` is GraphQL-only, so the suggested fix ("use `gh api repos/...`") is unsatisfiable for this operation. Working a PR review loop end-to-end (reply to a review thread, then resolve it) hard-stops at the resolve step; the REST reply works, the resolution needs a human click.

## Suggestion

Teach the guard a narrow GraphQL allowance: for a known-safe mutation set (`resolveReviewThread`, `unresolveReviewThread`, similar review-thread housekeeping), verify ownership from the query's input when it names a node the session already touched via verifiable REST calls — or at minimum, special-case mutations whose blast radius is review-thread metadata (no code, no refs, no releases) with a softer nudge instead of a hard block.

If the ruling is "working as intended — GraphQL stays blocked, ask the user," a doc line in the guard's block message naming the known GraphQL-only operations would stop the next session from hunting for a REST path that doesn't exist.

## Repro

Any `gh api graphql -f query='mutation { resolveReviewThread(input: {threadId: "..."}) { ... } }'` from a session with guard-gh-write active.

### LINKED ISSUE #317: feedback(guard-gh-write): blocks resolveReviewThread — no -R or REST form exists for gh api graphql
## What fires

`guard-gh-write` blocks resolving a CodeRabbit (or any) PR review thread. The only API for it is the GraphQL `resolveReviewThread` mutation:

```
gh api graphql -f query='mutation { resolveReviewThread(input:{threadId:"PRRT_…"}) { thread { isResolved } } }'
```

The guard rejects it two ways depending on shape:
1. In a loop → `gh write command in loop without explicit -R flag`
2. Individually → `gh api write to an unverifiable target — ownership can't be checked … Fix: use gh api repos/<owner>/<repo>/…`

Both fixes it suggests are impossible for this call:
- **`-R` doesn't exist for `gh api graphql`** — it's a raw GraphQL endpoint, not a repo-scoped subcommand.
- **There is no `repos/<owner>/<repo>/…` REST form** — GitHub exposes review-thread resolution *only* via GraphQL. The target is a global node ID (`PRRT_…`), which carries no owner/repo in the string for the guard to parse.

So a legitimate, routine post-review cleanup — resolving threads you've addressed — is unconditionally blocked, with no compliant path.

## Why it matters

Resolving addressed threads is the normal end of a CodeRabbit review loop (`cadence-forge:using-coderabbit` / `review-loop` both prescribe it). Today it can only be done by handing the mutation to the user via a `!`-prefix, or by leaving threads unresolved. The guard is protecting against unowned `gh api` *writes*, but a `resolveReviewThread` mutation is scoped to a node ID the session already obtained *from* the target PR — the ownership is transitively known even though it isn't in the argument string.

## Suggested shape (maintainer's call)

Recognize `gh api graphql` mutations whose body is a known-safe, PR-scoped operation (`resolveReviewThread`, `unresolveReviewThread`, `addPullRequestReviewThreadReply`) and allow them — or, more generally, allow a `gh api graphql` write when the node ID it targets was surfaced earlier in the same session from an allowed-owner repo. At minimum, the error text shouldn't suggest two fixes (`-R`, `repos/<owner>/<repo>/`) that are both structurally impossible for `gh api graphql`.

## Provenance

Hit 2026-07-17 resolving 4 addressed CodeRabbit threads on cameronsjo/forgectl#98 (env module). Session oaken-reed. Worked around by leaving threads addressed-but-unresolved (all had public code fixes or disposition replies).

cameronsjo:
The gh-write guard's GraphQL handling had two defects pulling in opposite directions: a read misclassified as a write (#263), and a class of legitimate, allowlisted-owner mutations with no satisfiable path at all (#262, #300, #317).

## Classifier hardening (#263)

The mutation detector ran a case-insensitive word match over the **raw** query — so `query { __type(name: "Mutation") … }` (an introspection *read*) blocked. String literals (`"…"`, `"""…"""`, honoring escapes) and `#` comments are now stripped before classification, and the keyword match is case-sensitive (`mutation` the operation keyword, not `Mutation` the type). Reads classify on what the query *does*, not what its strings mention.

## Bounded safe-mutation carve-out (#262, #300, #317)

`gh api graphql` has no `-R`/`repos/<owner>/<repo>` form, so the guard's remediation for review-thread mutations (`resolveReviewThread` — the CodeRabbit thread-resolution workflow) was structurally unsatisfiable: the target is a node ID that carries no owner to verify. The fix allows exactly two mutations — `resolveReviewThread` and `unresolveReviewThread` — pure boolean thread metadata, no content posting. `addPullRequestReviewThreadReply` is deliberately excluded (posts attacker-controllable text; REST replies remain owner-verified).

The extractor is fail-closed: multiple operations, unbalanced structure, aliases resolving to non-allowlisted fields, non-inline (`@file`) payloads — all return "not safe" and block. The subset rule is ALL root fields allowlisted, never "contains", so a composite `mutation { resolveReviewThread(...) deleteRepository(...) }` blocks.

The block message now states the reality instead of prescribing the impossible: no `-R` form exists; the two thread mutations are auto-allowed; anything else runs by the user directly. Loop-wrapped thread mutations still route through the loop-write path and stay blocked (deliberate; possible follow-up).

## Tests

33 GraphQL cases. The block matrix is the point: repository/ref/release/merge mutations, the excluded reply mutation, composite disguise, alias disguise, string-injection through literal braces, `@file` payloads, and a chained read-then-mutation. Allow side: both safe mutations (plain, multiline, aliased), the #263 introspection regression, and the existing read shapes.

Closes #262
Closes #263
Closes #300
Closes #317

### DIFF
--- a/crates/guardrails/src/guard_gh_write.rs
+++ b/crates/guardrails/src/guard_gh_write.rs
@@ -637,6 +637,23 @@ fn strip_graphql_literals(query: &str) -> String {
                 // Consume until the closing `"""` (or end — unterminated blanks
                 // the remainder, which fails the extractor closed).
                 while i < chars.len() {
+                    // GraphQL's ONLY block-string escape is `\"""` (a literal
+                    // triple-quote). It does NOT close the string — consume all
+                    // four chars and stay inside. Missing this let a smuggled
+                    // root field ride through: an early close turned the real
+                    // `}` chars living inside the string into structural braces,
+                    // so the extractor closed the selection set early and never
+                    // saw the trailing mutation field (#262).
+                    if chars[i] == '\\'
+                        && i + 3 < chars.len()
+                        && chars[i + 1] == '"'
+                        && chars[i + 2] == '"'
+                        && chars[i + 3] == '"'
+                    {
+                        out.push_str("    ");
+                        i += 4;
+                        continue;
+                    }
                     if chars[i] == '"'
                         && i + 2 < chars.len()
                         && chars[i + 1] == '"'
@@ -2905,6 +2922,59 @@ mod tests {
             );
             let result = GhWriteGuard.run(&input);
             assert!(matches!(result.outcome, cadence_hooks_core::Outcome::Block));
+        });
+    }
+
+    #[test]
+    fn graphql_block_string_escaped_triple_quote_rider_blocks() {
+        // SECURITY REGRESSION (#262): a `\"""` escaped triple-quote keeps the
+        // block string open past the `}` chars smuggled inside it. Without
+        // honoring the escape the stripper closed early, the real `}`s read as
+        // structure, the extractor closed the selection set before the trailing
+        // field, and `deleteRepository` rode through as a safe
+        // `resolveReviewThread`. Must BLOCK.
+        with_env(&owners_env(), || {
+            let input = input_with(
+                r#"gh api graphql -f query='mutation { resolveReviewThread(input:{threadId:"""X\"""} } } """}) {clientMutationId} deleteRepository(input:{repositoryId:"NODE_ID"}) {clientMutationId} }'"#,
+                "/tmp",
+            );
+            let result = GhWriteGuard.run(&input);
+            assert!(
+                matches!(result.outcome, cadence_hooks_core::Outcome::Block),
+                "smuggled deleteRepository must not ride through as a safe mutation"
+            );
+            assert_eq!(
+                result.block_metadata.expect("structured block").rule_id,
+                "gh-write-api-unverifiable"
+            );
+        });
+    }
+
+    #[test]
+    fn graphql_block_string_escaped_triple_quote_legit_allows() {
+        // A genuine `\"""` inside a string value, with no smuggled field, must
+        // still ALLOW — the escape is honored in both directions, not blanket-blocked.
+        with_env(&owners_env(), || {
+            let input = input_with(
+                r#"gh api graphql -f query='mutation { resolveReviewThread(input:{threadId:"""has \""" a literal triple-quote"""}) {clientMutationId} }'"#,
+                "/tmp",
+            );
+            let result = GhWriteGuard.run(&input);
+            assert!(matches!(result.outcome, cadence_hooks_core::Outcome::Allow));
+        });
+    }
+
+    #[test]
+    fn graphql_block_string_bare_brace_no_escape_allows() {
+        // A normal block string containing a bare `}` (no escape) with a single
+        // safe root field must still ALLOW — the escape fix didn't over-tighten.
+        with_env(&owners_env(), || {
+            let input = input_with(
+                r#"gh api graphql -f query='mutation { resolveReviewThread(input:{threadId:"""note } with brace"""}) {clientMutationId} }'"#,
+                "/tmp",
+            );
+            let result = GhWriteGuard.run(&input);
+            assert!(matches!(result.outcome, cadence_hooks_core::Outcome::Allow));
         });
     }
 
--- a/crates/guardrails/src/guard_gh_write.rs
+++ b/crates/guardrails/src/guard_gh_write.rs
@@ -58,10 +58,26 @@ static REPO_SUBCOMMAND: LazyLock<Regex> = LazyLock::new(|| {
 static API_REPOS: LazyLock<Regex> =
     LazyLock::new(|| Regex::new(r"/?repos/([^/]+/[^/ ]+)").expect("pattern should compile"));
 
-/// Word-boundary, case-insensitive match for the GraphQL `mutation` keyword —
-/// the signal that a `gh api graphql` query writes rather than reads.
-static MUTATION_WORD: LazyLock<Regex> =
-    LazyLock::new(|| Regex::new(r"(?i)\bmutation\b").expect("pattern should compile"));
+/// Word-boundary match for the lowercase GraphQL `mutation` operation keyword —
+/// the signal that a `gh api graphql` query writes rather than reads. Matched
+/// case-SENSITIVELY: the operation keyword is lowercase, whereas the *type* name
+/// `Mutation` (e.g. an introspection read `__type(name: "Mutation")`) is
+/// capitalized and must not be mistaken for a write (#263).
+static MUTATION_KEYWORD: LazyLock<Regex> =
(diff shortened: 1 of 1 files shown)


cameronsjo:
Merging after a two-lens review panel plus a fix cycle (in lieu of the rate-limited automated reviewer).

Security lens (adversarial bypass hunt) found a **Critical false-allow** in the first revision: the GraphQL block-string literal stripper did not honor the `\"""` escaped-triple-quote, so a crafted query could smuggle a second root field (e.g. `deleteRepository`) past the safe-mutation allowlist — it rode through as `resolveReviewThread`. Fixed in e328b94 (the block-string scanner now consumes `\"""` as an escape, matching GraphQL's lexer), with three regression tests including the exact rider (proven to fail without the fix, pass with it). Re-review confirmed CLOSED with no new divergence (double-escape, trailing-backslash, two-quote, and offset-desync cases all verified fail-closed).

Correctness lens: PASS — byte-walking extractor is UTF-8-safe, tests are adversarial (alias/composite/string-injection), fail-closed on ambiguity. Full guard suite 180/0 green (run independently), CI green both platforms.

PR #343, opened 2026-07-20
fix(enforce-worktree): separate-git-dir tightening, in-chain dismiss, coordination messaging, fixture hardening (#255 #313 #323 #312 #304)

### LINKED ISSUE #255: enforce-worktree: --separate-git-dir primary checkouts read as linked worktrees (false-allow)
## Observed (CodeRabbit finding on #253, deferred there by design)

`worktree::is_primary_checkout()` (crates/core, lifted unchanged from the guardrails crate in #253) classifies a checkout by whether `.git` is a directory. A primary checkout created with `git init --separate-git-dir <elsewhere>` has a `.git` **file**, so it reads as a linked worktree → `would_block_here()` is false → `enforce-worktree` never blocks there and the SessionStart posture line stays silent.

## Why it wasn't fixed in #253

- The semantics predate #253 — the predicate lift changed the *home* of the check, not its behavior, and the PR's mandate was narration-only (no block-decision changes; plan `docs/plans/2026-07-06-surface-the-invisible.md`).
- Line/wall **parity is preserved**: both callers share the one predicate, so the posture line and the guard agree in this configuration (both silent).
- The failure mode is a false-**allow** on an exotic setup, which is the friction-budget principle's preferred direction (false-allows over false-blocks).

## Ask

Decide whether `--separate-git-dir` primaries are worth supporting. If yes: classify via resolved `--git-dir` == `--git-common-dir` (equal → primary; a linked worktree's git-dir sits under `<common>/worktrees/<name>`) instead of the `.git`-is-dir heuristic, in the one shared predicate. Note the current heuristic is **pure/filesystem-only**; the replacement shells out to git — worth checking the call sites' tolerance for that (SessionStart runs once/session, but the guard runs per Edit/Write).

If no: document the carve-out next to the predicate and close.

### LINKED ISSUE #304: feedback(cadence-hooks): enforce-worktree keys to the session anchor, not the command's cwd — misfires cross-repo
## feedback(cadence-hooks): `enforce-worktree` keys to the session anchor, not the command's cwd — misfires cross-repo

**cadence-hooks:** 0.55.0
**Hook:** `guardrails enforce-worktree` (PreToolUse — "Block mutations in a primary checkout of a branch-mode repo")

### Problem

When a session legitimately runs a `git commit` inside a *second, independent* repository —
a separate clone with its own `.git`, nested under or sited beside the session's anchor
checkout — `enforce-worktree` evaluates the commit against the **session's anchor repo**
rather than the repository the command **actually runs in**. The commit is in a perfectly
valid location, just not the anchor's expected worktree, so the block **cannot be satisfied**
by entering a worktree of the repo being committed to. The only way past is a dismissal.

### Evidence (real occurrence)

Hit during a real multi-repo cleanup session: the primary checkout was anchored on repo A; a
routine `git commit` inside an independent repo B (a distinct clone kept inside A's tree so an
agent inherits the parent directory's context) was blocked by `enforce-worktree`, which named
A's expected worktree — unsatisfiable from B. Resolved only via the logged dismissal
(`dismiss-enforce-worktree --reason …`), which is a per-commit escape hatch, not a fix.

### Reproduction note

The misfire is **session-anchor-state dependent**, not argument-dependent, so a synthetic
sample payload does not reproduce it. Feeding a hand-built PreToolUse payload (a
`git commit -m test` command with `cwd` set to `/tmp/some-other-repo`) via:

```
$ cadence-hooks try guardrails enforce-worktree --payload p.json --show-payload
Outcome:  ALLOW (exit 0)
```

A cwd-only payload ALLOWs because it carries no anchor state and `/tmp/...` isn't a
branch-mode primary checkout. A faithful repro needs the real shape: a session anchored on
branch-mode repo A, then a `git commit` whose cwd is a *different* branch-mode repo B on its
own primary checkout — the guard blocks B's commit but keys the "expected worktree" it names
to A.

### Expected vs actual

- **Expected:** evaluate the mutation against the repo the command actually runs in — resolve
  the live toplevel at hook time (`git -C "$cwd" rev-parse --show-toplevel` /
  `--git-common-dir`) and check membership there.
- **Actual:** evaluates against the remembered session anchor, producing an unsatisfiable
  block whenever a command runs in a repo other than the one the session started in.

(Reporting the seam, not prescribing the internals — the general failure class is a check
keyed to a remembered anchor instead of the command's live cwd.)

### LINKED ISSUE #312: enforce-worktree Scratch test harness false-allows fixtures under .claude/ or /tmp (carve-out confound)
## Summary

`enforce-worktree`'s end-to-end test harness (`Scratch`) **silently false-allows every fixture** when the checkout sits under a location-based carve-out (`.claude/` or a temp dir). The block-expecting tests then fail on genuinely-correct code — a phantom regression that is a function of *test location*, not the code under test. This has bitten verification at least three times (#226, #253, and again during the #309 fix).

## Mechanism

`Scratch::new` anchors fixture repos relative to the checkout:

```rust
// crates/guardrails/src/enforce_worktree.rs (~L1693)
let root = Path::new(env!("CARGO_MANIFEST_DIR"))
    .join("../../target/enforce-worktree-scratch")
    .join(format!("{tag}-{}", std::process::id()));
```

When the checkout is a worktree under `.claude/worktrees/<slug>`, `CARGO_MANIFEST_DIR` resolves to `…/.claude/worktrees/<slug>/crates/guardrails`, so every fixture repo path carries a `.claude` component. The guard's own `is_claude_managed_dir` returns true if `.claude` appears **anywhere** in the path, so every fixture — meant to be a normal primary checkout that should *block* — is classified as Claude-managed and **allowed**. Every block-expecting test fails.

The harness is boxed between two location carve-outs by design: the `Scratch` comment notes it deliberately avoids `/tmp` (to dodge `is_temp_root`), but `target/` then inherits whatever prefix the checkout sits under — including `.claude/`. There is no fixed relative location free of both.

## Root asymmetry

`is_temp_root` is **injectable** — `assess_dir` threads `cfg.tmpdir`, so tests can pin it — but `is_claude_managed_dir` reads the path directly with no injection point. That asymmetry is why one carve-out is testable in isolation and the other silently contaminates fixtures.

## Proposed fixes (Layer A — the harness)

Either or both:

1. **Fail loudly at `Scratch::new`.** `assert!` that the resolved fixture root contains neither a `.claude` path component nor a temp prefix, with a message like *"fixture root sits under a carve-out (.claude/ or temp) — run the suite from a carve-out-free checkout."* This converts a silent false-pass into an immediate, self-explaining failure. Lowest-effort, highest-signal.
2. **Make the carve-out predicates injectable** the way `tmpdir` already is, so tests pin them off (mirrors the `with_worktree_env` discipline in `crates/session/src/start.rs`).

## Layer B (doc-only observation, not a behavior change)

The runtime `.claude`-anywhere match means a *primary* checkout literally located under a `.claude/` path component is exempt from the block. This is intentional for the two documented cases (`~/.claude/` config+memory tree; `.claude/worktrees/`, already exempt as worktrees). Real-world exposure is benign (the main primary-checkout-under-`.claude/` case is the plugin cache, which shouldn't be committed to anyway), so this warrants a comment noting the lexical match stands in for "is this actually Claude-managed state" — not a code change.

## Context

Surfaced while shipping #309 (`fix(enforce-worktree)`, v0.59.0). The `cadence-hooks/CLAUDE.md` gotcha already documents the *workaround* (run the suite from a carve-out-free path); this issue is to make the harness enforce it rather than relying on the operator remembering.

### LINKED ISSUE #313: feedback(cadence-hooks): enforce-worktree prescribes an impossible worktree when the target branch is already checked out in the primary
**Problem:** `guardrails enforce-worktree` blocks writes in a primary checkout and prescribes "create a worktree" — but when the write must land on a branch **already checked out in that primary** (a second session committing convention gap-fill onto a live peer session's feature branch), the prescribed remedy is impossible: git refuses to check out one branch in two worktrees (`fatal: '<branch>' is already used by worktree`). The only viable path is the dismiss, but the block text never acknowledges this case, so each session re-derives that conclusion from first principles.

**Version:** cadence-hooks 0.59.0

**Reproduction:**

Payload (`enforce-worktree-payload.json`):

```json
{"tool_name":"Write","tool_input":{"file_path":"/Users/cameron/Projects/artificer-design-system/themes/_research/lane1-294-terminalbg-ratification.md","content":"x"},"cwd":"/Users/cameron/Projects/artificer-design-system"}
```

```
$ cadence-hooks try guardrails enforce-worktree --payload enforce-worktree-payload.json
Outcome:  BLOCK (exit 2)
Stderr:
  Blocked: `/Users/cameron/Projects/artificer-design-system` is a primary checkout — feature work belongs in a worktree...
  Create one: `git worktree add .claude/worktrees/<slug> -b feat/<slug>` ...
  One-off exception: `cadence-hooks guardrails dismiss-enforce-worktree --for 30m` ...
```

**Live occurrence (2026-07-12):** session `velvet-anvil` in `artificer-design-system`, executing an owner-approved plan to land a docs-only commit (CHANGELOG + ratification packet) on `feat/terminal-raised-surface` — the branch a live peer session had checked out in the primary. Coordination discipline (declared lane, explicit-path adds, no branch switches) required committing on that exact branch in that exact checkout. Resolved via `dismiss-enforce-worktree --for 30m --reason "..."` (recorded in the bypass log), which worked as designed.

**Not a false positive per se** — the block is correct by its own rule; the defect is the remedy path. The guard has the inputs to detect the case: the write's target repo's `HEAD` branch vs. the branches of registered live sessions (the same registry `session status` reads), or minimally that `git worktree add` for the current branch would fail.

**Expected:** when the current branch cannot be duplicated into a worktree (already checked out here / peer-coordination case), the block text acknowledges it and points at the dismiss as the sanctioned path for that case.
**Actual:** the block prescribes an impossible worktree and leaves the dismiss reading like a workaround rather than the designed path.

**Generalizable, not one-off:** recurs whenever two coordinated sessions share one checkout and the second must commit onto the first's branch — a documented coordination pattern, not this session's odd shape.

### LINKED ISSUE #323: feedback(cadence-hooks): enforce-worktree blocks a chain that leads with its own dismiss command
## Problem

`enforce-worktree` evaluates the entire Bash command chain for mutating operations before the chain's own leading `dismiss-enforce-worktree` can execute. A single chain that *starts* with the sanctioned dismiss and then commits is blocked because the chain also contains `git commit` — the dismiss never gets to take effect. This forces the dismiss and the commit into two separate Bash calls every time.

## Environment

- `cadence-hooks 0.59.0`
- Hook: `guardrails enforce-worktree` (PreToolUse), interacting with `guardrails dismiss-enforce-worktree`
- Owner: `cameronsjo`, working in a branch-mode primary checkout

## Observed behavior

Command shape (single Bash call):

```
cd <repo> && cadence-hooks guardrails dismiss-enforce-worktree --for 30m --reason "..." && git add ... && git commit ...
```

The PreToolUse hook blocks the whole chain because it contains `git commit`, before the leading dismiss runs. Observed 5+ times in one session. Compounding friction: each 30m dismissal window also lapses during long verification waits, multiplying re-dismissals.

## Reproduction note

Could not trip the block in an available checkout: `cadence-hooks try guardrails enforce-worktree` returns ALLOW for the chain in a non-enforced checkout (`~/Projects/claude-configurations`), since the repo must be an actively-enforced branch-mode primary checkout for the guard to fire. The report is from live session observation in such a checkout; the payload shape is:

```json
{"tool_name":"Bash","tool_input":{"command":"cadence-hooks guardrails dismiss-enforce-worktree --for 30m --reason \"quick commit\" && git add -A && git commit -m \"chore: tweak\""},"cwd":"<branch-mode primary checkout>"}
```

## Expected vs actual

- **Expected:** a dismiss command at the head of the chain takes effect for the rest of that same chain.
- **Actual:** the whole chain is blocked on the presence of `git commit`, before the dismiss executes — two separate Bash calls required.

## Suggested seam

Either recognize a leading `dismiss-enforce-worktree` in the same chain (treat the rest of the chain as dismissed), or make the dismissal duration configurable / sticky per session so long verification waits don't lapse it mid-work.

cameronsjo:
Five commits across the enforce-worktree guard and its core predicate. One tightens a false-allow, two fix false-blocks, two harden the test fixtures and pin already-fixed behavior.

## `--separate-git-dir` primaries are primaries (#255)

`is_primary_checkout` classified by `.git`-is-dir alone, so a primary initialized with `--separate-git-dir` (whose `.git` is a file) read as a linked worktree — enforcement and the session-start posture line both silently skipped it. The predicate now resolves identity: primary iff canonical git-dir == canonical git-common-dir, spawn-free via the existing gitdir-file helpers. Genuine linked worktrees (gitdir under `…/worktrees/<name>`) still classify linked, and any resolution failure fails open — a broken `.git` never newly blocks.

## Block message acknowledges the can't-worktree case (#313)

When the change must land on the branch this checkout already has checked out (peer-coordinated work), `git worktree add` of that branch is mechanically impossible — yet the message prescribed exactly that and framed dismiss as a workaround. One added clause names the coordination case and states the dismiss is the sanctioned path there.

## Leading `&&`-chained in-chain dismiss (#323)

`dismiss-enforce-worktree … && git add … && git commit …` in one call blocked, because the snooze marker the dismiss writes doesn't exist when the guard evaluates — forcing two calls for a sanctioned sequence. The guard now recognizes a leading top-level dismiss and licenses a same-repo commit ordered after it, with strict scoping: top-level segments only (a dismiss inside `sh -c`/`$(…)` doesn't count), same repo only, dismiss-before-commit only, and **`&&` connectors only** — a `;`/`||` anywhere between them still blocks, so a dismiss that fails at runtime can never leave the commit executable. The allow carries synthesized Dismissal provenance, keeping the bypass log intact.

## Test-fixture hardening (#312) and attribution lock (#304)

The Scratch fixture root inherits the checkout's path prefix, so suites run from a checkout under `/tmp` or a `.claude/` path silently trip the temp/managed carve-outs and false-pass block-expecting tests. Both Scratch builders now assert a carve-out-free fixture root and fail with an explanatory message. #304's reported cross-repo misattribution no longer reproduces (fixed by the earlier commit-target/GitState rework); a new end-to-end test locks cwd-based attribution across nested repos so it stays fixed.

Closes #255
Closes #313
Closes #323
Closes #312
Closes #304

### DIFF
--- a/crates/core/src/worktree.rs
+++ b/crates/core/src/worktree.rs
@@ -31,11 +31,43 @@ pub fn is_truthy(value: Option<&str>) -> bool {
     )
 }
 
-/// Returns true if `repo_root` is a **primary checkout** — its `.git` is a
-/// directory. A linked worktree's `.git` is a *file* pointing into the
-/// primary repo's `.git/worktrees/`.
+/// Returns true if `repo_root` is a **primary checkout** — classified by
+/// resolved git identity, not by the surface form of `.git`. A conventional
+/// primary has a `.git` *directory* (the fast path). But a
+/// `--separate-git-dir` primary *also* has a `.git` *file* (like a linked
+/// worktree does), so the file-vs-dir surface form alone can't tell them apart
+/// (cadence-hooks#255): a separate-git-dir primary's `.git` file resolves to a
+/// git-dir that IS the shared common dir, whereas a linked worktree's `.git`
+/// file resolves to a git-dir under `.git/worktrees/<name>`, distinct from the
+/// common dir. So the rule is: **primary iff the canonical git-dir equals the
+/// canonical git-common-dir**, resolved the same spawn-free way
+/// [`crate::gitstate::GitState::resolve`] does.
+///
+/// Any resolution failure returns `false` (fail-open per ADR-0001 — a broken
+/// `.git` must never *newly* block).
 pub fn is_primary_checkout(repo_root: &str) -> bool {
-    Path::new(repo_root).join(".git").is_dir()
+    let root = Path::new(repo_root);
+    let dot_git = root.join(".git");
+    // Fast path: a real `.git` directory is always a primary checkout.
+    if dot_git.is_dir() {
+        return true;
+    }
+    // `.git` is a file (or absent): resolve the worktree admin dir (git-dir)
+    // and the shared common dir, then compare canonical identities. Reuses
+    // `crate::paths` exactly as `GitState::resolve` does — no git subprocess.
+    let Some(git_dir) = crate::paths::read_gitdir_file(&dot_git, root) else {
+        return false;
+    };
+    let Some(common_dir) = crate::paths::resolve_git_common_dir(root) else {
+        return false;
+    };
+    match (
+        std::fs::canonicalize(&git_dir),
+        std::fs::canonicalize(&common_dir),
+    ) {
+        (Ok(g), Ok(c)) => g == c,
+        _ => false,
+    }
 }
 
 /// Pure-ish: does `path` live under a temp root? `tmpdir` is the `$TMPDIR`
@@ -254,10 +286,75 @@ mod tests {
 
     #[test]
     fn not_primary_checkout_when_git_is_file() {
+        // A real linked worktree: its `.git` file resolves to a git-dir under
+        // `.git/worktrees/<name>`, distinct from the shared common dir — so the
+        // resolved-identity check classifies it as NOT primary.
+        let tmp = tempfile::tempdir().unwrap();
+        let primary = tmp.path().join("repo");
+        std::fs::create_dir_all(&primary).unwrap();
+        let git = |dir: &Path, args: &[&str]| {
+            let ok = std::process::Command::new("git")
+                .arg("-C")
+                .arg(dir)
+                .args(args)
+                .output()
+                .unwrap()
+                .status
+                .success();
+            assert!(ok, "git {args:?} failed");
+        };
+        git(&primary, &["init", "-q"]);
+        git(&primary, &["config", "user.email", "t@t"]);
+        git(&primary, &["config", "user.name", "t"]);
+        git(&primary, &["commit", "-q", "--allow-empty", "-m", "init"]);
+        let wt = tmp.path().join("wt");
+        git(
+            &primary,
+            &[
+                "worktree",
+                "add",
+                "-q",
+                &wt.to_string_lossy(),
+                "-b",
+                "feat/x",
+            ],
+        );
+        assert!(!is_primary_checkout(wt.to_str().unwrap()));
+        // Control: the primary checkout itself IS primary (`.git` is a dir).
+        assert!(is_primary_checkout(primary.to_str().unwrap()));
+    }
+
+    #[test]
+    fn separate_git_dir_primary_is_primary() {
+        // `git init --separate-git-dir` produces a `.git` *file* on a genuine
+        // primary checkout — its git-dir equals the common dir, so it must be
+        // classified primary despite the file surface form (cadence-hooks#255).
+        let tmp = tempfile::tempdir().unwrap();
+        let work = tmp.path().join("work");
+        let gitdir = tmp.path().join("gitdir");
+        std::fs::create_dir_all(&work).unwrap();
+        let ok = std::process::Command::new("git")
+            .arg("init")
+            .arg("-q")
+            .arg(format!("--separate-git-dir={}", gitdir.display()))
+            .arg(&work)
+            .output()
+            .unwrap()
+            .status
+            .success();
+        assert!(ok, "git init --separate-git-dir failed");
+        assert!(work.join(".git").is_file(), "sanity: .git is a file");
+        assert!(is_primary_checkout(work.to_str().unwrap()));
+    }
+
+    #[test]
+    fn separate_git_dir_resolution_failure_fails_open() {
+        // A dangling gitdir pointer (target does not exist) resolves to no real
+        // git dir — fail open to NOT primary rather than newly blocking.
         let dir = tempfile::tempdir().unwrap();
         std::fs::write(
             dir.path().join(".git"),
-            "gitdir: /repo/.git/worktrees/feat\n",
+            "gitdir: /nonexistent/dangling.git\n",
         )
         .unwrap();
         assert!(!is_primary_checkout(dir.path().to_str().unwrap()));
--- a/crates/core/src/worktree.rs
+++ b/crates/core/src/worktree.rs
@@ -124,6 +124,15 @@ fn normalized_components(dir: &Path) -> Vec<OsString> {
 
 /// Returns true if `dir` lives inside a `.claude/` directory — Claude Code
 /// tooling/state, never the branch-worthy product work these guards target.
+///
+/// Layer B: this lexical `.claude`-anywhere match is a coarse stand-in for "is
+/// Claude-managed state", not a precise repo classification. One consequence
+/// worth stating plainly — a *primary checkout* that itself lives under a
+/// `.claude/` path is intentionally exempt from enforce-worktree. That coarse
+/// exemption is exactly why the `#312` `Scratch` fixtures assert their roots
+/// stay OUT of `.claude/` (and out of any temp root): a fixture under a
+/// carve-out would silently exempt every case and make the block paths
+/// untestable.
 pub fn is_claude_managed_dir(dir: &Path) -> bool {
     normalized_components(dir)
         .iter()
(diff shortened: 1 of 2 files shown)


cameronsjo:
Merging on an independent two-lens review panel (in lieu of the rate-limited automated reviewer). Security lens (adversarial, guard-bypass hunt): CLEAN — every adversarial command string (cross-repo dismiss, all five non-`&&` connectors, dismiss-in-subshell, dismiss-after-commit, spoofed --repo/cd keying) traced to BLOCK; the is_primary_checkout fail-open path only triggers on a .git state where the commit itself can't land. Correctness lens: PASS spec-compliance — commit_target_of extraction verified verbatim/behavior-preserving, the &&-connector loop has no off-by-one, fixtures correct. Full suite run independently against this branch matches pristine main (the two known environment-dependent doctor tests only). One Important follow-up filed separately: gitstate.rs::worktree_git_dir still uses the old .git-file heuristic that #255 fixed in is_primary_checkout — a pre-existing gap this PR does not worsen (the enforce-worktree decision path and posture line are both fixed here).

PR #363, opened 2026-07-21
fix(metrics): resilient payload parse + remove dead log-plan-phase

### LINKED ISSUE #356: metrics: 234 stdin-parse failures in 7 days — identify the malformed feed
## Observed

`doctor` (2026-07-20 triage, v0.61.0-era): the metrics failopen log recorded **234 stdin-parse failures in the last 7 days**. The doctor's own threshold note says occasional malformed payloads are tolerated and 3+ suggests a wiring problem — 234/7d is two orders of magnitude past that.

## Read

Something is feeding the metrics logger malformed payloads at a steady rate — either a hook event whose payload shape changed under the parser, a wiring entry sending the wrong event to the logger, or a payload the parser never supported. The failopen design means zero user-visible symptoms while the data silently drops.

## Task

Sample the failopen log entries, identify which event/shape is failing to parse, and either fix the parser or the wiring. Worth checking against current real payload shapes — a payload-capture probe now ships with the hook-authoring guidance and showed several undocumented fields live (e.g. `effort`, `prompt_id`, `background_tasks`).

cameronsjo:
Two metrics-crate changes surfaced by a triage of the 234 stdin-parse-failure signal in `doctor`.

## 1. Resilient `tool_response`/`tool_input` parsing (#356)

The heartbeat logger — the sole `PostToolUse matcher:"*"` logger, so the only one exposed to every tool — was recording ~242 stdin-parse failures per 7 days. Root cause: `MetricsInput` and `HookInput` type `tool_response` as a strict Bash-shaped struct (`{stdout, stderr, answers}`), but a tool's `tool_response` shape varies — a plain string (`Read`), an array (`Glob`), a Bash-style object (`Bash`). Because `from_json` deserializes the whole payload at once, one mismatched field failed the **entire** parse and silently dropped the metrics row (fail-open by design → zero user-visible symptom).

Fix: a `lenient_option` deserializer applied to `tool_input`/`tool_response` in both structs — a present value whose shape doesn't match the target degrades to `None` instead of failing the payload. For enforcement (`HookInput`) this is also a robustness gain: a weird `tool_response` no longer blinds a guard that only needs `tool_input`.

TDD: RED tests assert string/array `tool_response` parses to `None` with the rest of the payload intact — both failed with `invalid type: string, expected struct ToolResponse` before, pass after.

**Empirical repro** (isolated `CADENCE_METRICS_DIR`, piped to the bare `session heartbeat`): Bash-shaped object → clean; string- and array-shaped → one parse-failure row each.

## 2. Remove dead `log-plan-phase` subcommand (cadence#506)

Claude Code's platform never delivers plan-mode tool events (`EnterPlanMode`/`ExitPlanMode`) to `PostToolUse` hooks, so `plan-phases.jsonl` recorded zero rows since it shipped. Removed the subcommand, its module, the unused `PLAN_PHASE_SCHEMA_VERSION` const, and all registry/dispatch/clap wiring. The plugin-side wiring is removed in cameronsjo/cadence#528; the live plan-approval signal is captured by `plan-links.jsonl` (persist-plan).

`make ci` green (fmt + clippy `-D warnings` + tests).

Closes #356


cameronsjo:
CodeRabbit is adaptive-limited (0 reviews landed on this head `0aac592`; the stated ETA did not clear). Merging on an internal adversarial review in lieu of CodeRabbit:

- **Security pass (guard-bypass lens)** — required here because the change touches guard-feeding payload parsing. Verdict: strictly safe. The lenient parser can only *widen* a guard's visibility of a runnable command, never narrow it (a shape-mismatched payload that previously failed the whole parse → fail-open Allow now degrades one field to `None` while siblings still parse → the guard runs). No payload exists where the old code would block and the new allows.
- **Code-quality pass** — two findings, both fixed in `0aac592`: the stale `docs/hooks.md` row was removed; the deliberate silent-degradation tradeoff is now documented, with the variance-vs-drift observability enhancement tracked in #364.

All required CI checks (Check ubuntu-latest, windows-latest) are green; CodeRabbit is a non-required check. Recording per the merge-over-a-limited-reviewer disposition.

PR #384, opened 2026-07-23
feat(rules): reject a plugin prefix in a skill's frontmatter name

cameronsjo:
## Why

Claude Code builds a skill's invocation id from `<plugin>:<directory>` and, as of **2.1.216** ("fixed plugin skills with a `name` frontmatter field losing their plugin prefix in slash-command autocomplete"), **prepends the prefix itself**. So a declared `name: cadence:tend` renders as `/cadence:cadence:tend` in the slash menu. **2.1.218** then made agent markdown reject `:` in a name outright, reserving the character for plugin namespacing.

This **reverses 0.19.0**, which relaxed `NAME_PATTERN` to permit the prefix after **2.1.94** made plugin skills use the frontmatter `name` as the invocation name — which is what made the prefixed form render correctly for four months.

## What changed

- `NAME_PATTERN` drops the optional `(:[a-z0-9]+(-[a-z0-9]+)*)?` group, so a colon fails the format check.
- The name-vs-directory check becomes a direct comparison. The `rsplit_once(':')` suffix extraction is **deleted** rather than left as an unreachable shim.
- The error message names the bare form instead of describing an optional prefix that can no longer be used.
- Tests flipped: `valid_name_format` now asserts the prefixed forms **fail**; `run_skill_namespaced_name_matching_dir_passes` became `run_skill_namespaced_name_blocks_even_when_suffix_matches`; a new `run_skill_bare_name_matching_dir_passes` covers the correct form. `run_skill_namespaced_name_suffix_mismatch_blocks` stays — still blocks, now for two reasons.

## Merge ordering

**Merge and release this AFTER cameronsjo/cadence#545, cameronsjo/cadence-lab#12, and cameronsjo/auditing-claude-md#5.** Those sweep 142 skills to bare names. Released first, this guard blocks every edit to a still-prefixed `SKILL.md`.

## The trap this documents

The comment block above `NAME_PATTERN` carries the order-of-operations hazard for a future flip, because this convention has now moved three times with the platform:

> If the platform flips back — e.g. Anthropic de-duplicates an already-prefixed name — **relax the pattern and ship a release FIRST, then sweep.** Tightened as it stands, this check blocks every edit to a prefixed `SKILL.md`, *including the sweep that would undo it.*

It also names exactly what to restore (the optional trailing group and the `rsplit_once` comparison), so the reversal is a lookup rather than an excavation. The sweep itself is `cadence/scripts/skill-names.py --bare|--prefixed`, added in cadence#545, which runs in either direction.

## Verification

- `cargo test -p cadence-hooks-rules` — **831 passed, 0 failed**
- `cargo fmt --all` + clippy `-D warnings` — clean (first pass caught a `taken reference of right operand` on the new comparison; fixed)
- **Live probe** against a real `skills/<dir>/SKILL.md` path, since the check no-ops and false-PASSes on paths without a `skills/<dir>/` ancestor:

| Payload | Expected | Actual |
|---|---|---|
| `name: tend` in `skills/tend/` | exit 0 | **exit 0** |
| `name: cadence:tend` in `skills/tend/` | exit 2 | **exit 2** — "must be the bare skill directory" |
| `name: wrong` in `skills/tend/` (control) | exit 2 | **exit 2** — "must match directory 'tend'" |

The control matters: it proves the validator engaged rather than no-opping on the path.

### One note on local `make ci`

5 `cadence-hooks-core::worktree` tests fail when `cargo` runs from a linked worktree outside `.claude/worktrees/`. They fail identically on **pristine `main` content** in the same location and pass in the primary checkout, so they're location-dependent, not a regression from this diff. CI has no such worktree and should be green.


cameronsjo:
Review note on a rationale this PR deletes.

The old comment justified permissiveness this way:

> the namespace itself is not verified against the plugin name — deriving the plugin from the path is fragile (source vs cache) and cadence-hooks must not force the prefix on non-cadence users.

That reasoning was sound and it does **not** carry over as an objection here, for a reason worth stating explicitly: rejecting the prefix isn't imposing a cadence convention on non-cadence authors — it's enforcing a **platform** requirement. Bare names are what Claude Code 2.1.216 requires of *every* plugin, cadence or not; a non-cadence author who declares `myplugin:myskill` gets `/myplugin:myplugin:myskill` exactly the same way.

So the guard is doing strictly less imposing than before: it no longer has an opinion about namespaces at all, it just requires `name` to be the directory. And the fragile part the old comment worried about — deriving the plugin from the path — is now unnecessary rather than merely skipped, since there's no namespace left to validate.

PR #387, opened 2026-07-23
feat(rules): accept background frontmatter field, enforce strict booleans

cameronsjo:
## What

Claude Code 2.1.218 made `context: fork` skills background-by-default, with `background: false` as the per-skill opt-out. `validate-skill-frontmatter`'s strict field allowlist rejected `background` as an unknown field, so the opt-out could not be written under the guard.

## Changes

- Add `background` to `VALID_FIELDS`.
- New boolean value-validation, one rule for all three boolean fields (`background`, `disable-model-invocation`, `user-invocable`): exactly `true` or `false`. The platform also accepts `yes/no/on/off/1/0` as of 2.1.218 — this repo's house style deliberately does not; one spelling keeps a skill corpus greppable.
- Strip trailing inline YAML comments before the boolean compare (`background: true  # note` is the boolean `true`; a `#` without preceding whitespace stays part of the value per YAML). Quoted values (`"true"`) remain blocked — unquoted is the one house spelling.
- Regression tests: `true`/`false` accepted (bare and with an inline comment); `yes`, `1`, and quoted spellings rejected with the house-strictness message; unknown fields still rejected; strip-helper boundary cases.

## Verification

- `make ci` green locally (fmt-check + clippy `-D warnings` + full suite), run from a carve-out-free worktree.
- Corpus grep across all local skill sources found zero existing non-`true/false` values on the three boolean fields — no migration needed.
- Sequencing proof: with the installed 0.64.0 binary, writing a skill file carrying `background: true` is blocked ("Unknown frontmatter field"); this change unblocks it while keeping value strictness.

## Summary by CodeRabbit

* **New Features**
  * Added support for additional skill frontmatter fields, including `background`, `agent`, `hooks`, and `paths`.
  * Boolean frontmatter values are now validated consistently and must use unquoted `true` or `false`.
  * Inline YAML comments are supported when validating boolean values.

* **Documentation**
  * Updated the changelog with the new frontmatter validation behavior.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
 ## [Unreleased]
 
+### Added
+
+- **`validate-skill-frontmatter` accepts the `background` field and enforces strict booleans.** Claude Code 2.1.218 made `context: fork` skills background-by-default with `background: false` as the per-skill opt-out; the field allowlist now admits `background` so the opt-out is writable. New value-validation applies one rule to all three boolean fields (`background`, `disable-model-invocation`, `user-invocable`): exactly `true` or `false`. The platform also accepts `yes/no/on/off/1/0` as of 2.1.218 — cadence deliberately does not; one spelling keeps the skill corpus greppable.
+
 ### Fixed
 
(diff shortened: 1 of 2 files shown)


cameronsjo:
@coderabbitai review

cameronsjo:
@coderabbitai full review

cameronsjo:
@coderabbitai full review

PR #414, opened 2026-07-25
fix(core): let a newline terminate a cd target; widen the ship anchor (#394, #368, #303)

### LINKED ISSUE #303: is_polish_ship_anchor: two pre-existing low-severity detection misses (sh -c wrapper, gh global-flag-before-subcommand)
## Summary

`core::shell::is_polish_ship_anchor` (the `nudge-polish-before-pr` / `log-polish-nudge` predicate, added in #297) has two low-severity detection misses. Both are **pre-existing** — inherited unchanged from the old `is_gh_pr_create` — and both cost only a **missed fail-open nudge**, never a guard bypass or false block. Surfaced by the `cadence-forge:security-reviewer` gate on #297; filed for a future hardening pass rather than fixed inline (the seam is nudge-only, no state mutation).

## The two misses

**L1 — shell-wrapper miss.** A ship command wrapped in a shell invocation is not seen:

```
sh -c 'gh pr create --title x'
bash -c "gh pr ready 12"
```

The predicate tokenizes the raw command, so the quoted script collapses to a single token and `gh`/`pr`/`create` never line up in the 3-window. `command_segments` (already in `core::shell`) expands `sh -c '…'` wrappers and would surface these, at the cost of also expanding substitutions.

**L2 — global-flag-before-subcommand miss.** A `gh` global flag before the subcommand breaks token adjacency:

```
gh --repo owner/r pr create --title x
```

Tokens are `[gh, --repo, owner/r, pr, create, …]`, so no `[gh, pr, create]` window matches. The normal form `gh pr create --repo x` matches fine.

## Why low

- **Fail-open, nudge-only** — worst case is a `/polish` nudge that doesn't fire; there is no block to bypass and no state mutation.
- **Pre-existing** — the old `split_whitespace`-based `is_gh_pr_create` missed both identically; #297's `tokenize` swap did not regress them (it only made the predicate *stricter*, killing quoted-phrase false positives).
- **Rare shapes** — nobody ships a PR through `sh -c`, and the `--repo`-before-subcommand form is unusual.

## Possible fix (deferred)

Evaluate the anchor over `command_segments(command)` instead of `split_segments(command)` to catch L1, and/or relax the adjacency check to tolerate global flags between `gh` and `pr`/`create`/`ready` for L2. Neither is urgent given the fail-open, nudge-only stakes.

## Provenance

Predicate: `crates/core/src/shell.rs::is_polish_ship_anchor` (#297, v0.57.0; segment-scoping fix v0.57.1).

### LINKED ISSUE #368: feedback(cadence-hooks): pre-PR polish gate misses record-polish marker in a linked worktree
## Problem

The pre-PR polish gate (`nudge-polish-before-pr`) fires "No polish recorded for this branch" on `gh pr create` **even after polish ran and recorded its marker this session** — when the session is working in a linked git worktree.

## Reproduction

- `cadence-hooks 0.64.0`
- Work in a linked worktree: `git worktree add .claude/worktrees/<slug> -b <branch>`.
- Run a full `cadence-forge:polish` pass, then `cadence-hooks cadence record-polish --scope full` — it **succeeds**, reporting a written marker:
  `recorded polish marker: $TMPDIR/cadence-hooks-<h>/polish-<repoHash>-<branchHash> (<repo>/.git@<branch> scope=full)`
- Then run `gh pr create -R <org>/<repo> ...`.
- The `nudge-polish-before-pr` PreToolUse nudge fires anyway.

## Evidence — likely keying mismatch in a worktree

In a linked worktree the two relevant git roots differ:

```
git-dir:    <repo>/.git/worktrees/<slug>
common-dir: <repo>/.git
branch:     <branch>
```

The marker filename is `polish-<repoHash>-<branchHash>`. If `record-polish` derives `<repoHash>` from one root (e.g. the worktree `git-dir`, or the `.git@branch` label shown in its output) while `nudge-polish-before-pr` derives its lookup key from another (e.g. `git-common-dir`, or a branch/base computed "vs origin/main" per the nudge text), the read misses the marker the write just made. Worktree context is the discriminator — the same flow on a primary checkout presumably matches.

## Expected vs actual

- **Expected:** after `record-polish` succeeds this session, the pre-PR gate finds the marker and stays silent.
- **Actual:** the nudge fires on `gh pr create` despite the recorded marker, in a linked worktree.

## Lane

Reactive — hook fired in error (false-positive nudge). Not the operator's mistake: polish genuinely ran and `record-polish` reported success before the PR create.

### LINKED ISSUE #394: feedback(record-polish): marker written from a worktree cwd doesn't satisfy the ship-gate read from that same cwd
## Observed (2026-07-24, forgectl worktrees)

Running `cadence-hooks cadence record-polish --scope code` from a git worktree's cwd (e.g. `/private/tmp/fgc-wt-arm90`, branch `feat/arm-codesign-gate`) succeeds and reports:

```
recorded polish marker: .../polish-<hash> (/Users/cameron/Projects/claude-configurations/forgectl/.git@feat/arm-codesign-gate scope=code)
```

i.e. the marker is keyed to the **common git dir** (`forgectl/.git@<branch>`). But immediately afterward, `gh pr create` run from the **same worktree cwd** still fires the `nudge-polish-before-pr` nudge:

```
No polish recorded for this branch — run /polish ... before this PR ships
```

So the writer and the ship-gate reader resolve the marker key differently even from an identical worktree cwd. Reproduced twice this session (PRs cameronsjo/forgectl#156 and #157) — record-polish said 'recorded', the gate said 'not recorded'.

## Impact
The nudge is non-blocking, so PRs still open — but it defeats the marker's purpose for worktree-based flows (e.g. orchestrating-issue-slates ships every PR from a `/tmp` worktree cwd), producing a false 'no polish' nudge on every worktree PR even when polish/record-polish ran.

## Likely area
The worktree's `.git` is a file pointing at `<common>/.git/worktrees/<name>`; record-polish keys to the common dir + branch, but the gate may resolve to the per-worktree gitdir (or vice-versa). Adjacent to the #324 'marker keyed to the tree that ran polish' item — this is the writer/reader key resolution not matching from the same cwd.

## Repro sketch
1. `git worktree add /tmp/wt -b feat/x origin/main`
2. edit + commit in /tmp/wt
3. `cd /tmp/wt && cadence-hooks cadence record-polish --scope code`  → 'recorded ... /repo/.git@feat/x'
4. `cd /tmp/wt && gh pr create ...`  → still nudges 'No polish recorded for this branch'

cameronsjo:
## What this fixes

Two issues (#394, #368) were filed as a writer/reader keying divergence in the polish marker — worktree gitdir vs common dir. **That diagnosis is refuted.** `record-polish` and `polish_marker_present` have both keyed on the canonicalized `git_common_dir` since 0.60.0, and the clean worktree case reproduces silent.

The real defect was in the reader's cwd resolver, and it was two characters wide. `CD_PATTERN`'s bare-path class was `[^ &;|]+` — excludes a space, but not a newline. So the shape that actually shipped:

```
cd <worktree>
gh pr create --title x
```

captured the target as `<worktree>\ngh`. No such directory exists, `GitState::resolve` returned `None`, and the gate nudged. The marker had been correct the whole time.

The class is now `[^\s&;|]+`. That is the entire fix.

## Why the newline is NOT a separator

An earlier revision also added `\n` to the separator alternation, which would additionally recognize a `cd` on its own line *after* an earlier command. That edit is deliberately reverted.

It creates no new bypass class — `^`, `&&`, `;`, and `||` all reach the identical steering today, confirmed by a three-way old/new/bash differential. But it converts a rarely-tripped mis-resolution into a routinely-tripped one:

- this repo's own docs carry **7** line-start `cd` occurrences against **1** `;`-spelled
- `\s*` means an indented `cd` inside a fenced code block matches too
- `cd ~/...` expands to a directory that really exists, so those become live re-points

The triggering shape is `gh pr create --body-file - <<'EOF'` with a shell snippet in the body — a routinely composed body. Widening the accidental-trigger surface of a primitive that three block-capable guards resolve through is the wrong trade for a nudge-only false positive.

## #303 — two ship-anchor detection misses

Also included, and independently cleared:

- **`sh -c '…'` wrapper.** A ship wrapped in a shell invocation was invisible because the quoted script collapses to one token. `is_polish_ship_anchor` now evaluates `command_segments`, which expands `sh|bash|zsh|dash -c`.
- **gh global flag before the subcommand.** `gh --repo owner/r pr create` broke the strict `[gh, pr, create]` adjacency window. Detection now walks forward from the `gh` command word, skipping `-`-prefixed tokens and consuming one extra token after a bare `--repo`/`-R`, before requiring a literal `pr`.

Both misses were pre-existing and fail-open — the cost was a missed nudge, never a bypass or a false block. Per-segment draft scoping survives both changes, so `sh -c 'gh pr create --draft'` is still correctly skipped, and every existing negative still rejects because the walk demands a literal `pr` token.

The anchor work was cleared separately: no unbounded loop, no index hazard, fuzzed clean over 4000 random token inputs, 40k x 40k token stress in 0.07s, nudge-only and fail-open.

## Known gap, recorded not closed

The anchor now reads `command_segments` while `parse_work_dir` remains a raw-string scan, so the predicate deciding *whether* a segment ships and the resolver deciding *where* it ships from do not see the same view. In `sh -c 'cd /x && gh pr create'` the ship is seen but the `cd` is not. That gap pre-dates this change; moving the anchor widens it, and closing it belongs with the resolver.

## Verification

- `cargo test --workspace --no-fail-fast`: **3281 passed, 0 failed**
- `cargo fmt --all -- --check`: clean
- `cargo clippy --workspace --all-targets -- -D warnings`: clean, 0 findings

Closes #394, closes #368, closes #303

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved command parsing for newline-separated directory changes and heredoc content.
  * Improved detection of pull request creation and readiness commands, including shell wrappers and repository flags.
  * Draft pull requests are correctly excluded from ship-anchor detection.
  * Improved working-directory resolution for linked worktrees.

* **Tests**
  * Added coverage for shell wrappers, repository flag variations, heredocs, newline handling, and worktree workflows.

* **Documentation**
  * Updated the unreleased changelog with these fixes.


cameronsjo:
@coderabbitai review

PR #415, opened 2026-07-25
fix(enforce-worktree): resolve every commit form that names a target tree (#378, #377)

### LINKED ISSUE #377: feedback(cadence-hooks): enforce-worktree guard appears to key off the orchestrator's cwd, not the dispatched subagent's own worktree
## Title

feedback(cadence-hooks): enforce-worktree guard appears to key off the orchestrator's cwd, not the dispatched subagent's own worktree

## Problem

During a multi-agent session — one orchestrating session dispatching several parallel subagents, each working in its own `git worktree` in a different repo — the enforce-worktree guard fired once for a subagent while it was running a read-only Snyk/diff verification command from inside its own worktree. The subagent reported that the block appeared to key off the *orchestrating* session's cwd (which was in a different repo, working a different PR concurrently) rather than the subagent's own actual cwd/worktree.

## Diagnosis

Nothing broke — the subagent proceeded and completed its task successfully. This is a proactive dogfood-friction report based on the subagent's own real-time observation during the run, not a constructed minimal repro. The subagent's own words: "the enforce-worktree hook misfired once on a read-only snyk/diff command (keyed off the session's tron-scans-site cwd, not my actual worktree) — flagging as guard feedback, not acted on."

## Root Cause

Not confirmed. Hypothesis: enforce-worktree's cwd resolution for a dispatched subagent may read the parent/orchestrating session's cwd or session-level state rather than the actual process/tool-call cwd of the subagent that triggered the check.

## Suggested Fix

Needs investigation — not a confirmed fix. If a maintainer wants to reproduce: dispatch two or more parallel subagents from an orchestrating session, each into its own `git worktree` in a different repo, and have one subagent run a command the enforce-worktree guard inspects (e.g. a Snyk scan or `git diff`). Check whether the guard's cwd resolution reads the subagent's own process cwd or the parent orchestrating session's.

## Labels

cadence-hooks

### LINKED ISSUE #378: enforce-worktree: `git -C <repo>` from an outside cwd bypasses the block `cd <repo> && git commit` triggers
**Guard:** `enforce-worktree` (cadence-hooks)

**Observed** (macOS; session flagged cadence-hooks version skew + 5 plugin warnings at start, so a mid-session hook reload is plausible):

Working from an outer cwd (the session's primary repo), committing into a *nested* primary checkout at `<nested>/deep-dives`:

- `cd <nested>/deep-dives && git commit -m ...` → **blocked**: "Blocked: `<nested>/deep-dives` is a primary checkout — feature work belongs in a worktree."
- `git -C "<nested>/deep-dives" commit -m ...` run from the outer cwd → **passed**, no block. This was the reliable path all session.

**Why it's confusing:** the block message says *"Judged against `<repo>` (the repo this command targets via cd/-C), not `<cwd>`."* That wording implies `-C <repo>` and `cd <repo>` are judged identically — but empirically `-C`-from-outside was not blocked while `cd`-in was.

**Also:** intra-session inconsistency — an early `git -C <repo> commit` succeeded, a later `cd <repo> && git commit` blocked (consistent with a hook reload between them).

**Expected:** either enforce consistently regardless of `cd` vs `-C`, or fix the message so it describes the actual trigger (shell cwd being inside a primary checkout), since the current wording claims `-C` parity that doesn't hold.

cameronsjo:
## What this fixes

`enforce-worktree` skipped every commit form that names a target tree by flag or env. From a linked worktree, five spellings of "commit into the primary checkout" were allowed:

- `git --work-tree=<primary> commit`
- `git --git-dir=<primary>/.git commit`
- both separate-value spellings
- a `GIT_DIR=… GIT_WORK_TREE=… git commit` env prefix

`git -C <primary> commit` and `cd <primary> && git commit` blocked correctly the whole time — **the issue as filed reported those as broken and is refuted.** The generalization to the flag and env forms is the real bypass. The two flags used to set an `ambiguous` flag that returned early "because the target tree cannot be cheaply resolved", but both name a directory that resolves like any other, so the skip was a miss dressed as a fail-open.

`--work-tree` and `--git-dir` are **both** emitted rather than ranked against each other, because they name two different things a commit mutates — the tree it reads, and the repository whose HEAD and index it advances. So `git --git-dir=<primary>/.git --work-tree=<elsewhere> commit` now blocks on the repo it really advances instead of being waved through on the work-tree alone.

#377 is also refuted: `enforce_worktree` resolves solely from the payload `cwd` and reads no session state, no parent-session state, and no process cwd. Its real defect was separate — the subprocess-mutation nudge fired on a file the command *creates* (`git diff > report.txt`), asserting "this command mutates tracked files" for a brand-new scratch file. Now skipped via a `symlink_metadata` probe with no git spawn.

## The Critical this round closes

A round-2 security review found the exclusion that keeps the guard from false-blocking a legitimate worktree commit was itself a bypass.

`is_linked_worktree_admin_dir` was a lexical check on an **un-canonicalized** path, and `Path::components()` preserves `..`. So `--git-dir=<primary>/.git/worktrees/..` — which *is* `<primary>/.git` — matched the exclusion, the git-dir target was dropped, the walk fell through to the session's worktree, and the guard allowed a commit that landed on the primary's branch. **Six spellings evaded.**

**The structural defect worth carrying past this repo:** `GitState::resolve` canonicalizes, so *assessment* was always immune to `..`/`.`/`//`/symlink spellings — but every decision made on the **raw target string** was not. That asymmetry is what was exploitable.

`lexical_normalize` folds `.`, `..`, `//`, and a trailing slash with no filesystem access, so it is safe on paths that do not exist and cannot surprise a caller by resolving `/var` to `/private/var`. It differs from real resolution only when a component is a symlink, and that difference runs in the safe direction: it can only cause a target to be KEPT rather than excluded, and a kept target is then assessed by `GitState`, which canonicalizes properly.

`normalize_target` wraps it with the existing `<repo>/.git` -> `<repo>` collapse, and is now applied to **every** emitted target including the fallback, and symmetrically to the dismiss map's other side (`dismiss --repo`).

## The Important

`inchain_dismissed_commits` re-derived commit targets in a parallel walk that had already drifted from `collect_targets` — it recursed into no child scripts and threaded no inherited env. So `dismiss && GIT_WORK_TREE=<p> sh -c 'git commit'` blocked despite the user's own dismiss. It now calls the real walk per segment, making parity structural instead of a comment asserting it.

The top-level-only rule for *arming* a dismiss is unchanged: commit detection recurses into wrappers, dismiss detection deliberately does not, so a dismiss buried in `sh -c '…'` still licenses nothing.

## Verification beyond the suite

A guard's tests ask what it decides *given* a target — never whether it SEES the target the shell will act on. So this was verified against the real binary and real bash.

**17-case probe, 17/17 pass.** Fixture under `~/wt` deliberately: `path_under_temp_root` hardcodes `/tmp` and `/private/tmp` as carve-outs no env var can undo, so a `/tmp` fixture is exempt from the guard and every verdict would read ALLOW. All six evading spellings block, plus two deeper `..` foldings past the admin dir. Three legitimate worktree spellings still allow — no false-block regression.

**Three-way differential**, because a probe that passes on both old and new proves nothing:

| binary | 6 evasions | controls | legitimate worktree spelling |
|---|---|---|---|
| `61dc79c` (pre-exclusion) | BLOCK | BLOCK | **false-BLOCK** |
| `cc58d42` (exclusion added) | **ALLOW — the Critical** | BLOCK | ALLOW |
| HEAD | **BLOCK** | BLOCK | **ALLOW** |

**Walk cost measured, not reasoned about.** The dismiss walk now pays the same recursion the block walk does. Against 500 plain segments, 500 env-prefixed segments, 100 depth-3 `sh -c` nests, and 200 dismiss+commit pairs: **0.020s at HEAD against 0.018s at `cc58d42`**. Constant factor; it is called once per invocation, not per target.

**Suite:** 3291 passed, 0 failed. `cargo fmt --check` and `clippy -D warnings` clean.

## Deliberate over-reaches, annotated at their sites

Both err toward seeing *more* of a command, since a miss is the dangerous direction for a block channel:

- a git env prefix is inherited into `$(…)`/backtick bodies even though bash applies an assignment prefix only *after* expansion
- `..` is folded lexically rather than resolved

Enumerated misses that remain, now stated in the module header rather than left implicit: a prefix's own option flags (`nice -n 10 …`, `env -i …`), `xargs`-fed commits, wrappers past `MAX_WRAPPER_DEPTH`, and a git env var set by a standalone `export` in an earlier segment.

Closes #378, closes #377

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved worktree enforcement for more Git target formats, environment settings, wrappers, and chained dismissal commands.
  * Prevented unnecessary mutation warnings for newly created scratch files.
  * Updated warnings to identify the resolved target path more accurately.

* **Security & Improvements**
  * Sanitized displayed paths and other values by removing control characters and limiting excessive length, making hook messages safer and easier to read.


cameronsjo:
## Parked — Windows fail-open blocks the merge

Ubuntu green, **windows-latest red: 26 failures, 23 of them `left: Allow` where `right: Block` is asserted.** The fail-open direction, including the simplest case (`git commit` in a primary, no flags). Those same tests pass on `main`.

Root cause is portability, not the security fix: the branch's new string-keyed target handling versus Windows path spelling. One instance is already fixed here — `lexical_normalize` collected into a `PathBuf`, which re-joins with the platform separator, so `/p/.git` came back `\p\.git` and stopped matching the un-normalized targets the dismiss map is keyed on. That cut the count but did not clear it.

Tracked as #418, with the two remaining suspects (Windows canonicalization's `\\?\` verbatim prefix; native separators in payload-derived targets) and a note that diagnosis needs Windows CI iteration — there is no local Windows to bisect against.

**The #378 Critical is independently verified closed on POSIX** and that evidence stands:

- 17-case probe against the real binary, fixture under a carve-out-free path: 17/17
- three-way differential — the six evasions were live at `cc58d42` and are closed at HEAD, with the legitimate worktree spellings still allowed
- walk cost measured at 0.020s against 0.018s pre-fix under pathological segment counts

Left open rather than closed so the security work and its verification trail stay reviewable. A guard that allows on one platform what it blocks on another is worse than one uniformly wrong — the operator cannot tell which they are running — so this does not ship half-fixed.

cameronsjo:
## The Windows job is green — the fail-open is closed

**993 passed, 0 failed on windows-latest**, zero `FAILED` lines in the run. Before this, 26 `enforce_worktree` tests failed on Windows only, every one of them `left: Allow, right: Block` — including `commit_in_primary_blocks_and_in_worktree_allows`, `dismiss_after_commit_still_blocks`, and `git_dir_flag_commit_into_primary_blocks`. The guard was permitting on Windows exactly what it blocks on macOS and Linux. Ubuntu stays green.

## Verified independently

Because "the tests pass now" has two explanations and only one of them is good:

- **The production code carries the fix**, not the fixtures — `lexical_normalize` folds drive-absolute paths, `looks_absolute` replaces the `starts_with('/')` check that read `C:\...` as relative, and `git_env_overrides` resolves the `GIT_WORK_TREE`/`GIT_DIR` forms that #377/#378 are about.
- **The one assertion that changed was checked before it changed.** The last commit updates a `#[cfg(windows)]` fixture from `C:\repo\wt` to `c:/repo/wt`. That test's own doc comment already said the target is normalized rather than passed through verbatim, so the assertion contradicted the comment above it. I confirmed the normalized form is the correct expectation before sanctioning it: every emitted target and every comparison base goes through `normalize_target`, so consistent folding is what makes case-insensitive matching work at all, and `git_commit_targets`' results are internal — no `format!` or message site consumes them, so lowercasing costs no user-facing fidelity. No production behavior was relaxed to make a test green.
- **Both closing references survive the merge-up:** `closingIssuesReferences` resolves to #377 and #378.

## One caveat on the checks

**CodeRabbit did not review this.** Its check reads `pass`, but the reason is "Review limit reached" — it reports green without a review having run. Merging on the strength of the real CI (this is a public repo, so both jobs genuinely execute) and the verification above, not on that check.

PR #445, opened 2026-07-26
feat: add Codex harness compatibility

cameronsjo:
Adds Codex input normalization, compatibility manifest/reporting, fail-closed security parsing, schema-v2 mixed-harness metrics, paired fixtures, and full route coverage. Refs cameronsjo/claude-configurations#438 and cameronsjo/claude-configurations#439.

## Pre-merge review

A full review pass ran before merge, reproducing each finding against a binary built from the branch rather than reading diffs. Two Critical issues and several Important ones were found and fixed here.

**Critical — the MCP tool-name rewrite was harness-agnostic and disabled `guard-browser-device` under Claude Code.** `HookInput::from_json` rewrote the `tool_name` of any payload beginning with `mcp__` to `Write`/`Edit`/`Read`/`Grep` by substring match, ungated by harness — so it ran on Claude payloads too. `GuardBrowserDevice` matches `mcp__claude-in-chrome__`, so once rewritten it returned `allow()` and did not even record its session marker. Four of the twenty-two Claude-in-Chrome tool names escaped the device-confirmation handshake, including `read_page`, the most common first browser call. `guard-browser-device` is in `PROTECTED_GUARDS` — the set `CADENCE_DISABLE` is explicitly forbidden to neuter — and this change neutered it for a subset of tools without touching it.

Fixed with the durable shape rather than a harness gate: `tool_name` is preserved verbatim and the normalized view moved to a separate field, so a guard chooses which it wants. Every call site was audited. A regression test asserts all twenty-two Claude-in-Chrome names still reach the guard as a block.

**Critical — `apply_patch` file deletion bypassed `guard-rm` while the manifest certified it as adapted.** A patch `*** Delete File:` operation normalizes to a synthetic `Edit` carrying `operation: "delete"` and no command. `GuardRm` opened with a command-or-allow early return, so it allowed unconditionally. Codex's `apply_patch` is a native deletion primitive Claude Code does not have — Claude deletes through Bash `rm`, which is why the guard only ever needed to read commands. Adding the harness added the primitive; the guard was never taught about it. Meanwhile `config/codex-compatibility.json` declared `guard-rm` as `"adapted"` with evidence naming a bare path and no test, so the shipped compatibility report told an auditor deletion was covered when it was not.

`GuardRm` now handles the command-less delete route through the same path triage, and the evidence field for a security-critical `adapted` row must reference a real test, asserted by parsing the named test out of the referenced file. That constraint is what would have caught this at CI time.

**Important, also fixed**

- The Codex hardening was gated solely on `CADENCE_HARNESS`, while payload normalization was unconditional — so a session where the variable was unset normalized payloads (looking healthy) while failing open on exactly the paths the branch existed to close. A payload-shape fallback now supplements the environment check; the sniff can only ever add strictness.
- A blank line inside an `apply_patch` update hunk was rejected as an invalid line prefix, hard-blocking every security-critical hook. A blank context line is legal in a unified diff and any producer trimming trailing whitespace emits one, so this would have been the first thing an operator hit.
- The compatibility test asserted a checked-in JSON agreed with itself and could not fail except by hand-editing a status downward.
- Denial logging moved inside the per-target loop, so one `apply_patch` with N operations wrote N rows per security-critical hook into an uncapped ledger, bounded only by patch size. Now one aggregated row carrying a target count.
- The Ask-to-Block conversion — a headline behavior of this branch — had no test coverage at all. It does now, as does the stdin-parse fail-closed branch.
- Pricing-branch logic duplicated verbatim across two metrics modules is now one shared method.
- The generated compatibility report had no CI freshness gate despite the generator supporting `--check`.

Full workspace suite green: 3,853 tests, 0 failures.

Session-Name: stone-beacon
Session-Id: e2e85b05-e25d-470d-92e1-970540d337f4
Model: claude-opus-5
Harness: claude-code 2.1.220
Machine: cf6e768835c7


cameronsjo (commented):
## Review of this PR

Reviewed by a six-lens review (security, Rust correctness, reuse/simplification) with every finding independently re-verified against the code before it reached this comment. Context worth stating plainly: at the time of review this PR had **zero** reviews — the automated reviewer was rate-limited at creation and the green check reflects that, not a completed pass.

Verification run locally on the branch: **3407 tests across 30 suites, 0 failed**; clippy clean under `-D warnings`; `fmt --check` clean; the committed compatibility report reproduces exactly from its generator.

### Two blockers, in my read

**1. The fail-closed mechanism never engages in a default Codex install.**

The whole security story here keys off `CADENCE_HARNESS=codex`, which the wrapper sets only when `CODEX_HOME` is non-empty. Two independent checks say that will not happen:

- `CODEX_HOME` is an *optional override*. Codex documents `~/.codex/config.toml` as the default. I ran the real binary with the variable unset and the child process saw it unset.
- The Codex plugin manifests declare `"capabilities": ["Skills"]` with no hooks key, so Codex never invokes these wrappers through the plugin mechanism at all.

So `codex_fail_closed`, the `Ask` to `Block` conversion, and the wrapper's exit-2 branches are all unreachable in practice. Codex *does* support hooks — its CLI exposes a hook-trust flag — so this is wireable; it just is not wired. The covering test sets `CODEX_HOME` explicitly in the Codex case and deletes it in the Claude case, so it encodes the assumption rather than testing it.

**2. The metrics directory relocation is unrelated scope with data consequences.**

`metrics_dir()` moves its default from the config dir to an XDG state path, with no migration. A read-only legacy fallback exists but is wired into 2 call sites, both reading only `commits.jsonl`, against ~20 call sites that use `metrics_dir()` directly. On a real machine that leaves 10 of 12 live ledgers — including a 5000-line denials stream — invisible to `doctor`, the staleness alarm, and every analysis path on first upgrade. This is not gated on Codex and hits every existing install. It reads like it belongs in its own PR with a migration step.

### Fixed in this branch

- **Codex detection unified.** `CADENCE_HARNESS` was read at three sites under two different rules — two case-sensitive, one not. `CADENCE_HARNESS=Codex` would tag the ledger `codex` while both security paths treated the run as Claude, failing **open** exactly when metrics claimed Codex was driving. All three now share one case-insensitive check with a pure resolver and unit tests; unset, empty, and lowercase behaviour is unchanged.
- **Test telemetry isolated.** Two spawn sites ran the binary without pinning the metrics dir, so the suite appended production-shaped rows to the operator's real ledger. Reproduced live during this review. Fixed and verified empirically.

### Worth addressing, not fixed here

- The compatibility manifest marks 10 hooks `adapted` all citing one generic evidence string, three of them security-critical, with no hook-specific test. The covering test only asserts the evidence field is non-empty.
- The new delete path in the vault trash guard has no `run()`-level test — the one new test calls the private helper directly, bypassing the integration glue entirely.
- The per-target aggregation loop always returns a result, so the effort-skip path can no longer propagate and its match arm is dead. Latent today (no shipped check overrides the skip), real the day one does.
- The Claude/Codex pricing branch is pasted verbatim into both loggers, reintroducing the drift the shared breakdown module exists to prevent.
- `clean_path` reimplements the canonical path normalizer, and its output is re-normalized downstream anyway.
- Aggregated multi-target messages concatenate every target's message regardless of which outcome won the merge, so an advisory note can end up glued to an unrelated block reason.

Schema back-compat checked separately and is **clean by construction**: the resume-point parser is unchanged and matches on field presence, and the new marker is prefix-validated, degrading to a full rescan rather than corrupting. No test pins that, though, unlike the established old-row convention elsewhere in the metrics crate.

Issue #452, opened 2026-07-26
polish anchor: a retargeted gh pr create/ready nudges about the cwd's branch, not the PR's


cameronsjo:
## Problem

`is_polish_ship_anchor` resolves the branch from the cwd, but a `gh pr create`/`gh pr ready` can be pointed at a **different repository** without leaving the cwd. When it is, the nudge fires about a branch that has nothing to do with the PR being created.

Measured against the released binary from a repo with no polish marker, every kill switch unset:

```
gh pr create --title x                       NUDGE   <- correct
GH_REPO=other/repo gh pr create --title x    NUDGE   <- wrong branch
GH_REPO=other/repo gh pr ready 12            NUDGE   <- wrong branch
```

`gh` does not error on this — verified live during the #325 review: it resolves the current local branch name against the override repo. So the command acts on `other/repo` while `polish_marker_present` looks up `(cwd's repo, cwd's branch)`.

Same class for the flag spellings: `gh --repo owner/r pr create`, `gh -Rowner/r pr create`, `gh pr create --repo=owner/r`.

## Why it is filed separately

#325 closed this for `gh pr merge`, which needed it structurally — a bare merge is only resolvable against the cwd at all if nothing retargets it, so `GhPrInvocation::retargeted` was built there and is already computed for **all three** subcommands. It is simply only consulted by the merge arm.

So the fix is plausibly a one-line extension. It is not bundled into #325 because:

- it is **pre-existing** — the shapes above nudge identically on the binary before that change, so bundling would have hidden a behavior change inside a feature PR
- `create`/`ready` have never inspected their operands, and teaching them to is a behavior change to two anchors that currently fire unconditionally
- the failure direction differs. For merge, anchoring a retargeted command was a *new* false nudge the feature would have introduced. Here it is a standing one, and silencing it also silences the legitimate case where someone creates a PR for the cwd's branch in a fork — worth thinking about before flipping.

## Impact

Nudge-only. Two consumers, `crates/cadence/src/nudge_polish_before_pr.rs` (Nudge) and `crates/metrics/src/log_polish_nudge.rs` (Logger); neither can block. The cost is a nudge naming the wrong branch, plus a `polish_nudges.jsonl` row attributed to a branch that never shipped — which matters more than usual right now, since that ledger is the denominator for the polish quiet-week measurement.

## Not closable

[...]

Issue #282, opened 2026-07-10
Guard: block writes creating plugin-root docs/ or scripts/ (belt-and-suspenders for shipping cruft)


cameronsjo:
## Problem

The cadence marketplace copies each entire `plugins/<name>/` directory into the plugin cache with no
files-allowlist, so any non-runtime artifact placed at a plugin root (`plugins/<name>/docs/` or
`plugins/<name>/scripts/`) ships ~49× (a full copy per marketplace commit). The
[doc-placement pass](https://github.com/cameronsjo/cadence/pull/371) evicted the existing cruft and
added two recurrence guards in the **cadence** repo:

- `.gitignore` `plugins/*/docs/` and `plugins/*/scripts/`
- a `plugin-runtime-only` CI check (`git ls-files ':(glob)plugins/*/docs/*' ':(glob)plugins/*/scripts/*'`)

Those catch it at `git add` time and in CI. What's missing is a **local, pre-commit** guard in the
`cadence-hooks` binary that blocks a Write/Edit creating a file under a plugin-root `docs/` or
`scripts/` — so the cruft can't even land in a local commit before CI rejects it.

## Proposal

A `PreToolUse` guard (Write/Edit/MultiEdit) that fails when the target path matches
`plugins/<name>/docs/**` or `plugins/<name>/scripts/**` (plugin-*root* only — never skill-nested
`skills/<skill>/scripts/`, which are runtime assets). Message points at the canonical placement scheme
(monorepo-root `docs/`/`scripts/`, or the vault for field reports).

Deferred from the placement pass under prose-now / guardrail-deferred: the CI check + gitignore cover
the shipping risk today; this is the belt-and-suspenders local enforcement.

### Notes
- Scope the pattern carefully — the `*`-crosses-`/` asymmetry bit the CI check (a bare glob matched
  ~20 legitimate `skills/<skill>/scripts/` files). The guard must match plugin-root only.
- Only fires inside the cadence monorepo (or any repo with a `plugins/*/` layout) — gate on repo
  identity or the marketplace manifest so it doesn't misfire elsewhere.

cameronsjo:
The guard-design plan (cameronsjo/cadence-hooks#482, merged 2026-09-07 as `dbbeed4d`) carries the two rulings for this issue, both from 2026-08-09:

[...]

Issue #275, opened 2026-07-10
configure: own the guardrails identity setup (CADENCE_ALLOWED_OWNERS/REPOS) currently done by the guardrails-init skill


cameronsjo:
## Problem

The `cadence-guardrails:guardrails-init` plugin skill's entire job is: run `gh api user`, write `CADENCE_ALLOWED_OWNERS` / `CADENCE_ALLOWED_REPOS` into `~/.claude/settings.json` (including legacy-key migration), then self-destruct. That is a deterministic config-write — exactly the territory of this binary's existing `configure` subcommand, which the skill currently duplicates from the outside.

## Proposed change

Teach `cadence-hooks configure` (or a `configure guardrails` sub-step) to:

1. Detect the GitHub identity (`gh api user`, with a no-`gh` fallback prompt)
2. Write/update `CADENCE_ALLOWED_OWNERS` / `CADENCE_ALLOWED_REPOS` in `~/.claude/settings.json`, migrating any legacy keys
3. Report what changed, idempotently — safe to re-run after updates

`doctor` should verify the wiring (keys present, owners non-empty) so drift is visible.

Plugin-side counterpart (thin the skill to an invoker of this subcommand, or retire it): cameronsjo/cadence#360 — filed together with this issue; the skill stays until this lands.

Note: `CADENCE_BYPASS=1` already exempts `configure`/`doctor`, so this works during maintenance windows like the rest of the CLI surface.

---
_From the 2026-07-10 skills-typing audit of the plugin ecosystem, judged against the cadence:writing-extensions mechanism rubric._

Issue #409, opened 2026-07-25
log-polish-nudge has never collected a row — the polish-gate efficacy denominator is empty


cameronsjo:
## Problem

`polish_nudges.jsonl` does not exist in either metrics directory on this machine,
and appears never to have been created — despite `log-polish-nudge` being
registered, wired, and functional. If that generalizes, the polish-gate efficacy
denominator has collected zero data for the lifetime of the check.

That file is described in its own module docs as "the deterministic denominator
for the polish-nudge efficacy" question. An empty denominator means the gate's
value has never been measurable.

## The binary is fine — verified

Fed a crafted `PostToolUse` payload directly (piped to the bare subcommand, not
through `try`, which ignores stdin), the logger works:

```
CADENCE_METRICS_DIR=<scratch> cadence-hooks metrics log-polish-nudge < payload.json
exit=0
```
```json
{"agentId":null,"branch":"main","markerPresent":false,"parentSessionId":null,
 "polished":false,"repo":"cadence-hooks","sessionId":"probe-polish-000",
 "transcriptPath":"/dev/null","ts":"2026-07-25T02:16:40Z"}
```

It created the file and wrote a well-formed row. So the defect is upstream of the
binary: the hook is not firing in production.

## Differential — the one variable that differs

Two loggers in the same plugin, same event, same matcher, same period:

| hook | `if:` filter | rows |
|---|---|---|
| `log-commit` | `Bash(*git commit*)` | **188** |
| `log-polish-nudge` | `Bash(*gh pr create*\|*gh pr ready*)` | **0** (file absent) |

The sibling collects fine. The only structural difference is the alternation in
the filter.

## Hypothesis — labelled as such, not established

This repo's own CLAUDE.md already records: *"hooks.json `if:` globs cannot match a
pipe character — `\|` is the alternation separator. For checks that fire on
pipelines, enumerate tool names in the filter and let the binary's pipe check do
the precision work."*

That note was written about a different check, but it fits here in either of two
readings, and both produce the same zero:

1. the alternation is not honored, so the filter never matches anything; or
2. the alternation is honored, but a subject command containing a pipe cannot be
   matched — and in practice almost every real invocation is piped
   (`gh pr create … | tail`, `… | head`), which is exactly how these commands get
   written when their output is being read back.

Reading 2 also predicts the sibling's success: `git commit` is typically run
unpiped, so `Bash(*git commit*)` matches.

I did not isolate which reading is correct — that needs a controlled probe of the
[...]

PR #482, opened 2026-07-28
plan(guards): #452, #282, #275 — design decisions before build

cameronsjo:
One decision-ready plan document covering three design items, each verified against `origin/main` (3b22ffb) in a clean worktree rather than taken from the issue bodies — and in two of the three the verification changed the recommendation. For #452, the filed "plausibly a one-line extension" and its refined form are both broken: `operands_are_flags_only` (`crates/core/src/shell.rs:419-449`) rejects any non-flag token *including a flag's value*, so it is a merge-shaped predicate that does not transfer — gating create/ready on it flips 17 of the 21 positive create/ready assertions in `shell.rs` and takes the create anchor dark, since nearly every real `gh pr create` carries `--title <value>`. The plan recommends splitting the retargeting field and suppressing only on the inline `GH_REPO=`/`GH_HOST=` channel (zero test cost — the `GH_REPO` assertions at `shell.rs:1912-1915` are merge-only), while noting that `gh pr merge` over-suppresses today on a redundant `-R <own repo>`. For #282 it confirms no such guard exists and lays out both undecided questions as options with failure modes, recommending a segment-index rule that reproduces `:(glob)` semantics plus manifest-based scoping, at nudge tier first. For #275 the premise understates the work by three capabilities and one governance collision: `configure` is a project-scoped, interactive, non-spawning writer of a single key whose `run()` never returns, and `src/main.rs:779-793` deliberately disables it under Claude Code.

**Ruling landed (2026-07-29):** #275's governance collision is settled — Claude may verify guardrails allowlist state doctor-side, but may never write it, at any scope, permanently. This resolves 4 of the plan's 12 open questions outright and promotes doctor-side verification (Option 3) to the plan's lead deliverable for #275, decoupled from any writer question. It also opens one new question — given human-only, does cadence#360 thin `guardrails-init` to an invoker or retire it — which the doc answers with a recommendation (thin, not retire) grounded in cadence#360's own 2026-07-27 owner ruling, reconciled with human-only via the CLI's existing Claude Code refusal. A shared tier policy (nudge first, block is a later separately-justified step) now covers all three guards in one place instead of each restating its own answer.

No implementation, no issue comments. The three issues stay open until the features ship, so this PR carries refs only and no closing keywords.

Refs cameronsjo/cadence-hooks#452, cameronsjo/cadence-hooks#282, cameronsjo/cadence-hooks#275, cameronsjo/cadence#360.

## Resolved (5 of 12)

- **#282 tier** — nudge first, per the shared tier policy.
- **#275 — may Claude write allowlist keys** — no, permanently human-only.
- **#275 — subcommand modelled on `migrate-config` vs a `configure` sub-step** — moot as a writer question.
- **#275 — ship doctor-side verification as a first slice, decoupled from the writer decision** — yes; now the lead deliverable.
- **#275 — may allowlist keys be written at project scope** — moot, no writing at any scope.

## Still open (7 of 12)

**#452 — polish-anchor retargeting**

1. For `gh pr create`/`gh pr ready`, should an inline `GH_REPO=`/`GH_HOST=` prefix suppress the nudge while a `--repo`/`-R` flag keeps nudging? *(Recommend yes.)*
2. Should `--repo`/`-R` be resolved against the cwd's remote so only a genuine mismatch suppresses, or is that more machinery than a nudge-only guard deserves? *(Recommend defer.)*
3. `gh pr merge -R <the cwd's own repo>` is suppressed today even though the branch is correct — fix it in a separate issue, or leave it?
4. `gh pr ready <n>` from an unrelated branch looks the marker up on the wrong branch: accept as a documented miss, or suppress numbered-ready only when it is also env-retargeted?

**#282 — plugin-root guard**

5. Scope the guard by marketplace manifest, repo identity, bare structure, or an opt-in env var? *(Recommend manifest, with a segment-test pre-filter.)*
6. Is the plugin cache in scope for this guard, a separate one, or out of scope?

**#275 — configure identity**

7. Given human-only, does cadence#360 thin `guardrails-init` to an invoker or retire it for a documented terminal command? *(Recommend thin, not retire — cadence#360 already carries a 2026-07-27 owner ruling to this effect, and it isn't in tension with human-only: the sub-step it would invoke already refuses under Claude Code regardless of caller.)*

---

Session-Name: tidal-viola
Model: claude-sonnet-5
Harness: claude-code 2.1.215
Machine: cf6e768835c7


cameronsjo:
**Two rulings.**

1. **The plugin-root guard identifies its targets from what the repo declares about itself** (the marketplace manifest), gated behind a cheap directory-segment pre-filter so the read only happens on a plausible hit. Rejected: folder-shape matching (false-positives on any unrelated plugin monorepo), hardcoded repo names (breaks on fork/rename), opt-in (invisible by default).
2. **The installed plugin cache is out of scope** for this guard — ships the source-repo guard sooner; the cache question can come back on its own.

The `GH_REPO=` env-prefix suppression question is still open pending more detail.

---
Ruled 2026-08-09 via decision slate `2026-08-09-parked-pr-rulings`.

cameronsjo:
**Ruling: no split — keep nudging on both forms.** An inline `GH_REPO=`/`GH_HOST=` prefix does *not* suppress the branch-mismatch nudge; it stays treated the same as an explicit `--repo`/`-R` flag, as it is today.

This declines the plan's own recommendation. The reasons:

- The split rests on a distributional argument plus a single test fixture — **no observed incident** was cited in either direction.
- The two failure modes are asymmetric in a way the plan's framing understates. Over-nudging is visible and cheap: one wrong nudge, one mis-attributed ledger row. Under-nudging is **invisible**, and it deflates the denominator of the #409 polish-adherence measurement — adherence would read as improving precisely because measurement stopped.
- The guard is nudge-only, so the cost of the conservative choice is very low.

Revisit if the nudge ledger ever produces real evidence that the env-prefix form precedes wrong-repo mistakes at a different rate than the flag form. Until then, no suppression.

---
Ruled 2026-08-09, follow-up round to decision slate `2026-08-09-parked-pr-rulings`.

cameronsjo (commented):
Spec-compliance verdict: PASS — the fix repoints the 1(c) supersession blockquote (docs/plans/2026-07-28-guard-design-decisions.md:223) from the removed "Rulings on the seven open questions" heading to `see "Still open" at the end`, and that heading exists at line 540 (`## Still open — 7 of the original 12, now all ruled`).

Plan-adherence sub-verdict: no `docs/plans/*.md` governs this diff (the diff IS a plan doc, not code under a plan) — this is a docs-only, single-file, mechanical fix to a prior review finding; plan-absent severity does not apply (falls under the docs-only/mechanical carve-out).

Verification performed:
- `git show 4266629 -- docs/plans/2026-07-28-guard-design-decisions.md` — confirmed the only change is the described repoint (plus a harmless line-join).
- Swept every `see "` and `§` cross-reference in the file against `grep -n '^#\+'` heading list: all resolve — `§2(c)`→line 344, `§3(d)`→line 499, all four `Still open` references→line 540. No occurrence of the old "Rulings on the seven open questions" phrase remains anywhere in the file.
- `wc -l` confirms 597 lines total, matching the PR's stated single new file.
- `markdownlint-cli2` run: 25 pre-existing issues (MD060 table style, MD013 line-length, MD024 duplicate headings by design across three parallel `#452`/`#282`/`#275` sections, MD036, MD029) — none introduced by this commit, none in scope per the task brief.

Critical: none.
Important: none.
Nit: none — the prior Important (stale heading quote) is resolved; no new issues introduced.

`gh pr view 482` confirms `headRefOid` is still `4266629c9904accbe9f4a3c55080e77057a4207f`.

PR #516, opened 2026-07-29
fix(fixtures): root the scratch fallback at XDG_CACHE_HOME so it is sweepable

### LINKED ISSUE #505: design: should the fixture fallback root live in $CARGO_HOME or $XDG_CACHE_HOME?
A design question deferred from #403, which is a fix rather than the place to relitigate a location choice.

## The situation

#403 gives test fixtures a fallback root for when the checkout itself sits under an enforce-worktree carve-out (a `/tmp` worktree, following the documented convention). The chain is: the checkout's own `target/`-rooted path, else `$CADENCE_HOOKS_TEST_SCRATCH_ROOT`, else `$CARGO_HOME` (or `$HOME/.cargo`).

`$CARGO_HOME` was chosen for a property — it definitionally exists on any machine that can run `cargo test` at all, and it is not checkout-relative. That property is genuine.

## The concern

It was picked for that property, not for purpose, and the purpose argues elsewhere.

The primary root is under `target/`, which carries an implicit contract: **`cargo clean` is the sweep**. The fallback breaks that contract. `Scratch`'s `Drop` reclaims only the `{tag}-{pid}` leaf, so `~/.cargo/cadence-hooks-test-scratch/` persists indefinitely, and `cargo clean` never touches it. A test-fixture directory living permanently inside another tool's home directory is a side effect outside the repo on a contributor's machine.

`$XDG_CACHE_HOME` (falling back to `~/.cache/cadence-hooks`) is the purpose-fit location: it is where a tool is expected to leave cache-shaped residue, it is swept by the conventions users already have for that directory, and it is equally checkout-independent.

## What this issue is **not**

The pollution observed during #403's review — `~/.cargo/cadence-hooks-test-scratch` created on a checkout under no carve-out — was caused by two tests that passed hardcoded carve-out roots and therefore relocated on every run. **That is a test defect and is fixed in #403 itself.** This issue is only about where the fallback should live for machines that genuinely need it.

## What to decide

1. Move the fallback to `$XDG_CACHE_HOME` / `~/.cache/cadence-hooks`, or keep `$CARGO_HOME` and accept an empty directory left behind on machines that actually relocate.
2. If it moves: does the chain keep `$CARGO_HOME` as a further fallback, or is `~/.cache` plus the existing `$CADENCE_HOOKS_TEST_SCRATCH_ROOT` hatch sufficient?

Either answer is defensible. The reason to settle it deliberately rather than by default is that the location becomes a documented contributor-facing contract the moment `CONTRIBUTING.md` describes it — which #403 does.

## Constraint on any change

Whatever wins must stay **lexically checkable without touching the filesystem**. The resolver's `escapes_carveout` predicate is purely path-textual by design, so the carve-out guarantee holds without a probe write per candidate on a code path that runs from ~120 fixture call sites.

cameronsjo:
Closes #505

Moves the git test-fixture scratch root's third-tier fallback from `$CARGO_HOME` to `$XDG_CACHE_HOME`, under a `cadence-hooks/test-scratch` path.

## Why

`Scratch::new`'s `Drop` reclaims only the `{tag}-{pid}` leaf, so the root directory itself persists indefinitely. Rooted at `$CARGO_HOME`, that leaves permanent residue in another tool's home directory that `cargo clean` never sweeps — cargo has no knowledge of the path and no reason to touch it.

`$XDG_CACHE_HOME` is purpose-fit for disposable data and is swept by ordinary cache conventions. The property that motivated `$CARGO_HOME` in the first place — "exists on any machine that can build this crate" — is preserved, since both variables have computable defaults (`$HOME/.cargo`, `$HOME/.cache`).

## Shape

`xdg_cache_home()` / `xdg_cache_home_from()` mirror the existing `user_home()` / `resolve_home()` split in `crates/core/src/paths.rs` — the same injectable-function pattern already used for testability, not a new one.

The path also gains structure: flat `cadence-hooks-test-scratch` becomes `cadence-hooks/test-scratch`, so the tool owns one cache directory rather than scattering top-level entries.

## Verification

```
cargo test --workspace --no-fail-fast          exit 0   3788 passed / 0 failed
cargo clippy --workspace --all-targets -- -D warnings   exit 0
cargo fmt --all -- --check                     exit 0
```

Six new tests cover the resolver: env-var-preferred, `$HOME` fallback when the var is unset, `None` when neither is available, both carve-out redirect paths, and a should-panic case for a carved-out `$XDG_CACHE_HOME` with no override.

All three gates were run from this branch's worktree by the orchestrator, not taken from the implementer's self-report.

## Note on Windows

This repo's CI builds Windows with `-D warnings`, and a prior PR failed there twice — once on a `#[cfg(unix)]`-only import that was unused elsewhere, once on Unix path literals fed into a carve-out predicate. Both classes were checked for here, but Windows CI is the authority; if it goes red, that is the likely cause.

Session-Name: tidal-viola
Model: claude-opus-5
Harness: claude-code 2.1.215
Machine: cf6e768835c7

## Summary by CodeRabbit

- **Documentation**
  - Updated worktree testing guidance to explain fixture relocation to the XDG cache directory.
  - Clarified how relocation notices are emitted and how to configure fixture paths in sandboxed environments.

- **Bug Fixes**
  - Improved test fixture handling when temporary directories overlap the repository’s build directory.
  - Added clearer fallback behavior and error messages when no suitable cache location is available.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -112,8 +112,9 @@ instead, so the guard's block/allow logic is still exercised for real
 once per test binary (`cargo test --workspace` runs several, each its own
 process), not once per fixture.
 
-If your machine has neither a usable `$XDG_CACHE_HOME` nor `$HOME` (a
-stripped-down sandbox), point fixtures at an explicit directory instead:
+If your machine has neither a usable `$XDG_CACHE_HOME` nor any resolvable home
+directory (a stripped-down sandbox), point fixtures at an explicit directory
+instead:
 
 ```bash
 export CADENCE_HOOKS_TEST_SCRATCH_ROOT=/some/carve-out-free/dir
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -106,14 +106,14 @@ diagnostics that have no value on trivial sessions.
 worktree docs recommend) works fine for `cargo test --workspace`: fixture
 directories that would otherwise land under the checkout's own `target/` —
 itself under `/tmp` in that layout, which `enforce-worktree`'s carve-out
-exempts — automatically relocate to `$CARGO_HOME` (or `$HOME/.cargo`) instead,
-so the guard's block/allow logic is still exercised for real
+exempts — automatically relocate to `$XDG_CACHE_HOME` (or `$HOME/.cache`)
+instead, so the guard's block/allow logic is still exercised for real
 (cadence-hooks#403). A one-line `note:` on stderr says when this happens —
 once per test binary (`cargo test --workspace` runs several, each its own
 process), not once per fixture.
 
-If your machine has neither a usable `$CARGO_HOME` nor `$HOME` (a stripped-down
-sandbox), point fixtures at an explicit directory instead:
+If your machine has neither a usable `$XDG_CACHE_HOME` nor `$HOME` (a
+stripped-down sandbox), point fixtures at an explicit directory instead:
 
 ```bash
 export CADENCE_HOOKS_TEST_SCRATCH_ROOT=/some/carve-out-free/dir
--- a/crates/core/src/git_fixtures.rs
+++ b/crates/core/src/git_fixtures.rs
@@ -302,13 +302,27 @@ fn xdg_cache_home() -> Option<PathBuf> {
 /// names neither `.claude/` nor a temp root, so `escapes_carveout` would
 /// bless it and this fixture would be created relative to the test binary's
 /// cwd instead of failing loudly.
+///
+/// An empty value is only the *narrowest* spelling of that hazard, though:
+/// any relative value (`XDG_CACHE_HOME=cache`) lands in the same place, for
+/// the same reason. So the rule is absoluteness, not non-emptiness — which is
+/// also what the XDG base-directory spec requires, where a relative value is
+/// to be ignored (#516 review).
+///
+/// Each candidate is filtered on its own rather than the chain being filtered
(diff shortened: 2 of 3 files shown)


cameronsjo:
## Merge gate: a reviewer agent stood in for CodeRabbit

Recording this so the trail is not misread later.

CodeRabbit is rate-limited across this org — sustained, not a clearing countdown. It cleared once, reviewed a single PR elsewhere, and immediately re-limited. Its review on this PR sits at `aded0d41`, which predates the relative-`$XDG_CACHE_HOME` fix, so it is stale by its own request.

The documented disposition for sustained exhaustion is to substitute an independent code-review pass at the same bar: no Critical or Important findings, or it does not merge.

**Verdict: MERGE-CLEAN.** Four focus areas checked — the absoluteness filter validates both candidates independently rather than filtering after the choice; Windows paths are platform-correct and both CI legs are green at head; `cargo_home_from` is genuinely dead with zero remaining references; the carve-out check is untouched and now only ever sees absolute candidates.

One check worth naming, because it is the kind that usually gets skipped: the review confirmed the new tests are **not** a false green from an uncompiled file. `git_fixtures.rs` is gated behind the `test-builders` feature, which has no `default`, so green CI could have meant it never built. Five workspace crates enable it as a dev-dependency, and Cargo workspace feature unification therefore compiles and runs it on both platforms.

All three gates were additionally run against the branch worktree independently of the implementer's report: 3788 passed / 0 failed, clippy clean, fmt clean.

PR #517, opened 2026-07-29
fix(guard): split the concurrency fallback so a real zero stops counting every session

### LINKED ISSUE #512: warn_subagent_concurrency: the session-scoping fallback counts every session when the current one has no rows yet
Re-filed from #352, which was closed because the mechanism it named never existed. This is the narrow defect that most likely produced the symptom it reported.

## The defect

`count_live_subagents(jsonl, session_id)` (`warn_subagent_concurrency.rs:87`) is session-scoped — but the scoping is **conditional**. At `:79-80`, it applies only when at least one record in the file carries the current sessionId. When none does, it falls back to counting **every** record.

That fallback is exactly the **first-spawn** case: a session that has not yet written a subagent record sees the whole file, including every other session's rows.

From the outside this is indistinguishable from "the guard counts globally" — which is how #352 came to describe it that way.

## Why the fallback presumably exists

Almost certainly to avoid reporting zero when the schema or the writer changes and rows stop carrying a recognizable sessionId. Counting something is better than silently counting nothing. That intent is sound; the cost is that the degraded mode is also the **normal** mode for the first spawn of every session.

## What to decide

The fix is not simply "remove the fallback" — that trades an over-count for an under-count, and an under-count on a concurrency *guard* is the direction that fails open.

Options worth weighing:

1. **Distinguish "no rows for this session" from "no rows recognizable at all."** An empty-but-well-formed file means zero live subagents for this session, which is a real answer, not a degraded one. Only an unparseable or schema-mismatched file should trigger the fallback.
2. **Keep the fallback but mark the count as untrusted** in the nudge text, so an inflated number reads as "could not scope" rather than as fact.

## Verification bar

Both directions, or the fix proves nothing:

- A session with **no** rows of its own, in a file containing another session's rows, reports **its own** count — not the file total.
- A file that genuinely cannot be scoped still errs toward reporting rather than silence.
- A session with rows of its own is unaffected.

Prior art on the reported symptom: #352 observed 26 live against a real 4. Whether that reproduces under this mechanism is the first thing to measure — the observation was never verified, only the mechanism was refuted.

cameronsjo:
Closes #512

Splits the subagent-concurrency guard's session-scoping fallback so a real zero stops being reported as every other session's agent count.

## The defect

`count_live_subagents` collapsed two different states into one branch. When no record matched the current session id, it counted **all** sessions' live agents. But "no match" has two causes that demand opposite answers:

- **Well-formed log, zero rows for this session** — the session simply hasn't spawned anything yet. The correct count is **0**.
- **No session-scoping data anywhere** (empty, garbage, or a schema with no `sessionId` field) — genuinely unscopable. Counting all is the safer estimate.

Only the second deserves the fallback. The first was an over-count on every session's first spawn.

## Why not the alternatives

**Removing the fallback outright** trades an over-count for an under-count — and an under-count on a concurrency guard **fails open**, permitting exactly what it exists to flag. **Marking the number untrusted in the nudge text** relabels the bug without fixing it. The split is the only option that stops the over-count without ever under-counting.

## Shape

A `SessionScope` enum (`Matched` / `WellFormedNoMatch` / `Unscopable`) resolved by a pure `resolve_session_scope`, so the three-way distinction is named in the type rather than implied by a boolean. `count_live_subagents` stays pure and I/O-free.

## The fail-open door that was checked, not assumed

The original module doc justified the fallback partly on this: *"the parent-session stamping is a platform detail that may not line up."* If that were true — if a session's own agents were logged under a **parent** session's id — then the log would carry session data, none of it matching, and the new `WellFormedNoMatch → 0` path would under-count. Same fail-open direction, different door.

So it was measured rather than reasoned about. Against the live metrics log (4207 records across 20+ sessions), the orchestrating session's own id appears **1464 times — more than any other session's**. Stamping aligns; agents are not filed under a parent id. The concern does not reproduce, and the split is safe.

## Verification

```
cargo test --workspace --no-fail-fast    exit 0    3790 passed / 0 failed
```

Four tests hold the verification bar:

1. `session_filter_counts_only_matching_session` — rows exist for this session, only those count.
2. `well_formed_log_with_no_rows_for_this_session_reports_real_zero` — **the bug**, and it carries a **positive control**: the same input is asserted to return 2 when scoping is dropped, proving the test can discriminate rather than passing vacuously.
3. `schema_missing_session_id_field_is_unscopable_falls_back_to_count_all` — the fallback still fires where it should.
4. `mixed_valid_and_garbage_lines_trusts_the_valid_rows_scoping` — garbage lines cannot launder an over-count back in by tipping a well-formed log into "unscopable".

The suite was run from this branch's worktree by the orchestrator, and each of the four tests was confirmed present in the output by name — with a negative control proving the search could return nothing. A passing suite that silently skipped the new tests reads identically to one that ran them.

## Follow-ups filed separately

Probing the live log surfaced two unrelated issues, filed rather than folded in: 88 records carrying a literal `test` session id in the production metrics file, and the log growing unbounded while being fully parsed on every hook fire.

Session-Name: tidal-viola
Model: claude-opus-5
Harness: claude-code 2.1.215
Machine: cf6e768835c7

### DIFF
--- a/crates/guardrails/src/warn_subagent_concurrency.rs
+++ b/crates/guardrails/src/warn_subagent_concurrency.rs
@@ -13,10 +13,18 @@
 //! `SubagentStart` and a matching `SubagentStop`, each carrying `event`,
 //! `agentId`, and `sessionId`. So the live count is the number of distinct
 //! `agentId`s that have a `SubagentStart` record and **no** matching
-//! `SubagentStop`. Records are scoped to the current session by `sessionId`; if
-//! no record matches the current session id (the parent-session stamping is a
-//! platform detail that may not line up), the count falls back to **all** live
-//! agents — the broader, safer estimate.
+//! `SubagentStop`. Records are scoped to the current session by `sessionId`;
+//! when at least one record carries the current session id, only matching
+//! records are considered.
+//!
+//! Two different "no match" cases must NOT collapse into one branch (#512):
+//! when the log is well-formed (some record carries *a* sessionId) but none
+//! carries *this* session's id, that's a real zero — the session just hasn't
+//! spawned anything yet — and the count must report 0, not every other
+//! session's agents. Only when the log carries **no** session-scoping data at
+//! all (empty, garbage, or a schema missing the `sessionId` field entirely) is
+//! the session genuinely unscopable, and the count falls back to **all** live
+//! agents — the broader, safer estimate for that case.
 //!
 //! FAIL-OPEN (ADR-0001): any failure to read or parse the log yields `allow()`.
 //! A missing log, garbage bytes, or an unreadable directory never blocks or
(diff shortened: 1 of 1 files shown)


cameronsjo:
## Merge gate: an independent security pass stood in for CodeRabbit

CodeRabbit has been rate-limited across this org for the duration — sustained, not a clearing countdown. The documented disposition is an independent review at the same bar, and for guard code a security lens is the floor regardless of what else ran.

**VERDICT: MERGE-CLEAN.** No Critical, no Important.

The review's most useful structural observation: `Matched` and `Unscopable` both reproduce the old behavior exactly, so **every** behavior delta from this change is confined to inputs classifying `WellFormedNoMatch`. That bounds the regression surface to one branch rather than the whole function.

## It found a reachable under-count — and it is still a net improvement

`warn_subagent_concurrency.rs:161`. `WellFormedNoMatch => false` assumes a session with no rows has no live agents. **On a session resume the id rolls**, and the harness re-stamps live agents under the new id only after a delay — during which the session genuinely has live agents and zero rows.

Measured from the live log rather than hypothesized, and independently re-verified before merging:

- Session `b114f4e0` started four agents at 14:31:04 / 14:31:24 / 14:31:45 / 14:33:58 on 2026-07-26.
- All four `agentId`s re-emit `SubagentStart` under `e1c08994` at **exactly 14:34:04Z** — one instant for all four, so a batch replay rather than four re-dispatches.
- Continuation direction is established, not assumed: `e1c08994` references `b114f4e0` 85 times, the reverse zero times.
- The gap between the predecessor's last transcript write (14:32:18) and the replay (14:34:04) is **~106 seconds**, during which a dispatch would see rows for 82 other sessions and none for itself → `WellFormedNoMatch` → 0, with four heavy agents live.

Arithmetic reproduced in isolation: the fixed code returns 0 where the old returned 5.

**Why this does not block.** The impact is a missed *advisory* nudge — the module's own contract is that the dispatch proceeds regardless. And the code being replaced was wrong on **every** session's first dispatch, counting 10 phantom live agents across 82 stale sessions. Trading a constant spurious nudge for a ~100-second silent window is strictly better.

A mitigation exists in data already logged: rows carry `cwd`, and a resumed session keeps its cwd, so `WellFormedNoMatch` could fall back to same-`cwd` live agents rather than 0. `parentSessionId` is null in every row checked, so the platform's own linkage is unavailable today. Filed as follow-up rather than folded in.

## A second finding worth recording

`has_any_session_data` is a whole-log predicate over an append-only file that grows without bound. All 4253 rows currently carry a `sessionId`, so on any machine with history the predicate is **permanently true** and `Unscopable` fires only on a genuinely fresh log.

Consequence: if the harness ever stops stamping `session_id`, new null-`sessionId` rows would not match, historical rows would keep the predicate true, and the guard would return 0 forever instead of taking the count-all fallback this PR deliberately preserved. The fallback is fresh-log-only in practice. Filed.

## Answering the adversarial question that was asked

Garbage lines **cannot** tip the classification toward the unsafe branch — they never parse into a record, so they contribute neither a match nor scoping data. The mixed-case test's reasoning is correct.

A *well-formed* appended line can: a schemaless log with 6 live agents for the current session returns 6, and appending one line carrying any `sessionId` returns 0. The precondition is a log with no `sessionId` anywhere, which the finding above shows is unreachable today, and the file is local and user-owned — accidents, not adversaries.

## Verified clean

- **No panic or blowup introduced.** No `unwrap`/`expect`/indexing/arithmetic outside tests. A 100,000-deep nested JSON line hits serde_json's recursion limit, returns `Err`, and is skipped — no stack overflow. Linear perf: 200,000 lines in 68 ms.
- **`resolve_session_scope` is pure** — no env reads, no I/O, no statics, no interior mutability.
- **Reader and writer resolve the same path.** The guard's inline `metrics_dir()` and `metrics::common::metrics_dir_from` are identical. Stated because a divergence here would now silently *zero* the guard where it previously over-counted.
- **Merge is clean.** `git merge-tree --write-tree` reports no conflict against `origin/main`; #516 and #521 touch no file this PR touches.
- **Test 2's positive control is genuine, mutation-verified** — flipping `WellFormedNoMatch => true` moves the assertion 0 → 2, so it fails on the regression it guards. Two weaknesses noted and filed: it establishes the input's liveness via the `Unscopable` path rather than the branch under test, and it duplicates an assertion another test already makes.

## Gaps, stated rather than implied

`semgrep`, `gitleaks`, and `cargo-audit` are all absent from PATH — no SAST or dependency-advisory pass ran.

Two things could not be settled:

1. **Whether `/clear` also rolls the session id, and whether the replay happens on that path.** `/clear` is far more frequent than resume, so if it rolls the id *without* a replay, the window widens from ~100 seconds to permanent for any agent surviving the clear. This is the open question that matters most; it needs an actual clear with live agents to answer.
2. **Whether a dispatch can land inside the roll window in practice.** The window is measured; no dispatch was observed inside one.

Orchestrator ran the suite independently at 3790 passing / 0 failed, and confirmed all four target tests present in the output by name with a negative control proving the search could return nothing.

Issue #520, opened 2026-07-29
prevent-secret-leaks: a cd inside a -c script never invalidates the .envrc carve-out, so the guard proves the wrong file


cameronsjo:
Found during the security review of the #508 fix. **Pre-existing — not introduced by #508** — but that fix widens its reach, so it should be settled alongside it rather than after.

## The split

Two different views of the same command disagree, and the carve-out trusts the weaker one.

- **The operand scan** uses `command_segments` (`crates/cadence/src/prevent_secret_leaks.rs:772`), which **recurses into `bash -c` bodies** and surfaces `cat .envrc` as its own segment.
- **The cd-invalidation** is computed by `command_changes_directory(&lower)` (`:751` → `:657`), which uses the **non-expanding** `split_segments`.

On `bash -c 'cd /secrets; cat .envrc'` the non-expanding splitter sees a single segment headed `bash` — the `;` is inside quotes, so it never splits — and reports no directory change.

## The consequence

`envrc_bash_read_allowed(".envrc", cwd, /* has_cd */ false)` (`:683`) joins the relative operand to `input.cwd`, reads the **project's own** `.envrc`, proves that file is a pure loader, and allows.

Meanwhile the shell has chdir'd and reads `/secrets/.envrc`, printing it into context.

The guard proved a property of one file and permitted the reading of a different one. That is precisely the wrong-file class #308 exists to prevent, reached through a wrapper the #308 check does not follow.

## Why #508 makes this more urgent without causing it

The verdict is identical before #508 — the old lowercased path expanded `bash -c` the same way and recovered the same token. So this is not a regression.

What changes is **reach**. Once #508 lands, `sudo -A`, `-E`, `-H`, and `-P` wrappers expand where they previously refused. Those spellings inherit this hole along with the fix's new blocks. The fix is still correct and should ship; this just stops being reachable only through unwrapped `bash -c`.

## Remediation

Compute the cd-invalidation from the **same expanded view** the operand scan uses — track it per-segment across `command_segments`, or recurse `child_scripts` with a fresh scope as `crates/core/src/shell.rs:2100-2106` already prescribes — rather than from a whole-command `split_segments`.

[...]

PR #521, opened 2026-07-29
fix(guard): segment before case-folding, closing four sudo-flag bypasses in prevent-secret-leaks

### LINKED ISSUE #508: prevent-secret-leaks lowercases before segmenting, so sudo -E hides the inner command from it alone
Found by an adversarial bash-vs-parser differential run during the #490/#491/#496/#497 segmenter work. **Five guards block this payload; `prevent-secret-leaks` alone allows it.**

```
sudo -E bash -c 'cat .env'
```

`guard-rm`, `guard-gh-write`, `git-safety`, `guard-op-vault-scan`, and `guard-gh-dangerous` all see through the wrapper and block. `prevent-secret-leaks` does not.

## Mechanism

`prevent_secret_leaks.rs:777` calls `command.to_lowercase()` **before** segmenting. By the time `skip_sudo_no_argument_flags` sees the argv, `-E` has become `-e`, which is not in the no-argument allowlist — so the walk refuses, the wrapper is never expanded, and the inner `cat .env` is never inspected.

The refusal itself is correct behavior (the allowlist is deliberately conservative: refuse on anything unrecognized rather than guess). The bug is that the input was mutated upstream into something the allowlist could not recognize.

## Why the obvious fix is wrong

**Do not case-fold the sudo allowlist.** Two independent reasons, both already recorded in this repo:

1. **#503's own commit message** documents that folding past the verb onto flags is exactly what regressed `-C`/`-P`/`-S` in #489. The rule that shipped was *fold the verb, and only the verb*.
2. **sudo's grammar is case-load-bearing.** `-P` takes no argument; `-p` takes a prompt string. `-C` takes a file descriptor; `-c` does not exist. A case-insensitive allowlist would accept `-p` as no-argument and then treat the prompt string as the command — turning a miss into a *misparse*, which is worse.

So this is not "add `eq_ignore_ascii_case`." The fix has to happen at the lowercasing site, not the matching site.

## Shape of the real fix

The question is why `bash_leaks_secrets` lowercases the whole command up front at all. Options, in rough order of preference:

1. **Segment first, lowercase later.** Run `command_segments` on the original string and apply case folding only where the guard actually compares verbs or filenames. This preserves flag case for the wrapper walk, which is the only consumer that needs it.
2. **Lowercase a copy for matching, keep the original for segmentation.** Cheaper to implement, but leaves two parallel representations in flight — the kind of divergence that produced this bug.
3. Leave it, and document that `prevent-secret-leaks` is weaker than its siblings on wrapper-prefixed commands. Not recommended; a guard that disagrees with five others on the same payload is a trap for whoever reads one of them and assumes the rest match.

Option 1 needs someone to check what else depends on the early lowercase — there is a documented interaction with `METADATA_SAFE_COMMANDS`, an **exemption** rather than a detector, where the upstream lowercase is currently load-bearing (`verb_fold_cannot_widen_this_guards_exemption` pins exactly that). Changing the lowercase site without accounting for that exemption could **widen** it, which is the one direction that subtracts blocks rather than adding them.

## Verification bar

Whatever lands must carry both directions:

- `sudo -E bash -c 'cat .env'` blocks — and the same payload with each of `-H`, `-n`, `--` blocks too.
- `sudo -u/bin/bash -c 'cat .env'` still does **not** expand. That is #493's discriminating negative control; if it goes green-to-red the fix has weakened the never-guess property.
- `LS .env` still receives `ls`'s exemption, and `XARGS echo < .env` still blocks — the two rows that pin the exemption boundary.

A test that only shows the new payload blocking cannot distinguish a correct fix from one that widened the exemption.

cameronsjo:
Closes #508

`prevent-secret-leaks` lowercased the command **before** segmenting it, so the shared parser never saw the original argv. `sudo -E bash -c 'cat .env'` reached this guard unsegmented while five other guards blocked it.

## Wider than filed, and in a different crate than filed

Two corrections to the issue, both verified against `main`:

**The path was wrong.** The guard is `crates/cadence/src/prevent_secret_leaks.rs`. The issue cites `crates/guardrails/`, where the file does not exist.

**Four flags bypass, not one.** `SUDO_NO_ARGUMENT_SHORT_FLAGS` (`crates/core/src/shell.rs:2314`) is `"AbEHiknPSs"` — case-sensitive, with its no-argument entries uppercase. After folding:

| flag | folds to | still allowlisted? |
|---|---|---|
| `-A` | `-a` | no — bypass |
| `-E` | `-e` | no — bypass |
| `-H` | `-h` | no — bypass |
| `-P` | `-p` | no — bypass |
| `-S` | `-s` | **yes**, coincidentally — lowercase `s` is a separate member |

The issue names only `-E`. `-S` survives by accident rather than design, and is pinned by a test so it stays that way for a reason.

## The fix already existed one file over

`bash_leaks_secrets` now calls `command_segments(command)` on the original, matching `prevent_secret_writes.rs:166`, which has always segmented the original while lowercasing only for its own comparisons. This is adopting the sibling's shape, not inventing one.

**The allowlist is deliberately untouched.** #503 records that folding past the verb regressed `-C`/`-P`/`-S` in #489, and sudo's grammar is case-load-bearing — `-P` takes no argument, `-p` takes a prompt string. A case-insensitive allowlist cannot represent that.

`original_case_token_at` and its cursor are removed as dead: once segments carry real case, `segment_env_reads` returns the true-case operand directly and there is nothing to recover.

## Tests were written first, and genuinely failed

Each bypass test was run against the **unfixed** code before the change. All four returned Allow. That is what makes them discriminating controls rather than assertions of convenience — a test authored after a fix passes on a parser wrong in the same way its author is wrong, which is exactly how two Criticals shipped through green suites earlier in this file's history.

```
sudo_capital_a_bash_c_cat_env_now_blocks          failed before, passes now
sudo_capital_e_bash_c_cat_env_now_blocks          failed before, passes now
sudo_capital_h_bash_c_cat_env_now_blocks          failed before, passes now
sudo_capital_p_bash_c_cat_env_now_blocks          failed before, passes now
sudo_capital_s_bash_c_cat_env_still_blocks        passed before and after (the coincidental survivor)
env_chdir_flag_exemption_survives_segment_before_fold   the narrowing direction
```

That last one matters for a reason the issue did not raise. The existing `verb_fold_cannot_widen_this_guards_exemption` pins the exemption against getting **wider**; nothing pinned it against getting **narrower**, and a silently narrowing exemption is a false-block regression — the opposite failure class, equally real. It now exercises `peel_env_options`'s case-insensitive matching with genuinely un-lowered input for the first time.

## A premise in #503's own PR description is now false

#503 asserts the verb fold is "a provable no-op [in this file] because `bash_leaks_secrets` already lowercases upstream." This change removes that upstream lowering, so the justification no longer holds — even though the conclusion still does, since `fold_verb` is an unconditional ASCII fold that behaves identically on pre-lowered or raw input.

Three doc comments were rewritten rather than left to mislead, because each was a written invitation to reopen a hole:

- `peel_env_options` — its case-insensitivity is now **load-bearing rather than compensatory**, and the two callers disagree about case (`command_dumps_env` still pre-lowers; `env_prefix_len` no longer does). Simplifying it back to uppercase-only would silently break `env -C /tmp`.
- `find_exec_leak`'s `EXEC_FLAGS` — case-sensitivity is deliberate, matching find's own predicate grammar. `find -EXEC` errors as an unknown predicate before executing anything, so an uppercase spelling is not a bypass.
- `command_word`'s cross-reference in `shell.rs`, for the same reason.

## Security review

An independent security pass found **no BLOCK→ALLOW regression**, tracing every case-sensitive comparison downstream of the change. Head resolution, operand classification, and `peel_env_options` all fold at their own sites. The heredoc and `$VAR` paths become *more* accurate — the old lowering collapsed `<<EOF`/`eof` and `$F`/`$f`, manufacturing blocks bash would never earn.

It also confirmed the `.envrc` carve-out is strictly tighter without `original_case_token_at`.

**Stated rather than implied:** `semgrep`, `gitleaks`, and `cargo-audit` were all absent from PATH, so no scanner evidence backs that pass. It is hand-traced plus the measured controls above.

## One finding filed separately, not fixed here

**#520** — a `cd` inside a `-c` script never invalidates the `.envrc` carve-out, because the cd check uses a non-expanding splitter while the operand scan expands. Pre-existing and not a regression, but this change widens its reach: the four sudo spellings now expand where they previously refused, inheriting the hole along with the fix.

It is filed rather than folded in because it is a different defect with its own verification bar. Same root shape as the earlier Round 2 Critical though — **two parsers with different rules applied to one command will disagree.**

## Verification

```
cargo test --workspace --no-fail-fast                    exit 0   3794 passed / 0 failed
cargo clippy --workspace --all-targets -- -D warnings    exit 0
cargo fmt --all -- --check                               exit 0
```

Run by the orchestrator against the branch worktree at `e188f99`, not taken from the implementer's report. The six new tests were enumerated **from the diff** and each confirmed present in the run output by name — a passing suite that silently skipped them reads identically to one that ran them.

Session-Name: tidal-viola
Model: claude-opus-5
Harness: claude-code 2.1.215
Machine: cf6e768835c7


cameronsjo:
## Merge gate: two independent review passes stood in for CodeRabbit

CodeRabbit is rate-limited across this org — sustained rather than a clearing countdown. The documented disposition is an independent review at the same bar: no Critical or Important findings, and for guard code a security pass is a floor regardless of what else ran.

Both ran here.

**Security pass — no BLOCK→ALLOW regression.** Traced every case-sensitive comparison downstream of the change: head resolution folds unconditionally, operand classification folds per-token, and the `env` flag arm is genuinely case-insensitive. The heredoc and `$VAR` paths become *more* accurate — the old lowering collapsed `<<EOF`/`eof` and `$F`/`$f`, manufacturing blocks bash would never earn.

**Correctness pass — MERGE-CLEAN.** Notably, it did not take this PR's positive-control claims on faith: it built a throwaway worktree at pre-fix `origin/main` and ran all four bypass payloads directly, confirming each returns Allow before the fix and Block after. That is independent reproduction rather than a reading of the test names.

It also settled the non-ASCII question as **moot rather than merely unaddressed**. The removed `original_case_token_at` needed a byte-length guard precisely because it mapped a position in the lowercased string back into the original — the one place Unicode length divergence could bite. Nothing does that mapping now: segmentation runs on the original directly, and every remaining lowercase scan is a self-contained boolean that is never cross-referenced by position. The design no longer has two parallel position spaces to diverge.

Conflict check against `origin/main` moving under this branch (#516 landed first): zero file overlap.

## One correction to this PR's description

The body above says `-S` "survives by accident rather than design." That is overstated, and the review caught it.

Both `-S` and `-s` are legitimately argument-free in real sudo's grammar — stdin-password and shell respectively. So the allowlist containing both is independently correct, not lucky. What is true is narrower: `-S` is unaffected by the fold *because* the allowlist is correct for both spellings, where `-A`/`-E`/`-H`/`-P` had no lowercase counterpart to land on.

The test pins observed Block behavior against the real guard rather than the folding mechanism, so a future allowlist edit that broke it would fail loudly. No silent-breakage risk either way.

## Gates, run independently of the implementer's report

```
cargo test --workspace --no-fail-fast    exit 0    3794 passed / 0 failed
cargo clippy --workspace --all-targets -- -D warnings    exit 0
cargo fmt --all -- --check               exit 0
```

The six new tests were enumerated from the diff and each confirmed present in the run output by name — a passing suite that silently skipped them reads identically to one that ran them.

**Not fixed here, filed instead:** #520 — a `cd` inside a `-c` script never invalidates the `.envrc` carve-out, because the cd check uses a non-expanding splitter while the operand scan expands. Pre-existing, but this change widens its reach.

Issue #657, opened 2026-08-08
guard-rm: .git component match is byte-exact — /tmp/repo/.GIT evades GitRoot on case-insensitive volumes


cameronsjo:
## Problem

`has_git_component` (guard_rm, and its pathclass twin) matches `seg == ".git"` byte-exact, and the git-root probe stats `<target>/.git`. On a case-insensitive volume (macOS default), `rm -rf /tmp/repo/.GIT` removes the real `.git` while both checks miss — the target classifies by location instead, which under the temp carve-out means Allow.

Pre-existing property of the whole GitRoot class (not introduced by the #576 fix), but the guard already folds case on the verb side (`command_word`, #488) — the two halves of the same guard currently disagree about volume case-sensitivity.

## Fix sketch

Case-fold the `.git` component match (and the probe where feasible) in the strict direction only: folding may promote a target into GitRoot protection, never demote one out of it.

Surfaced by the security review on the guard-rm cluster PR (#576/#569/#541).

Issue #428, opened 2026-07-25
guard-rm: cd tracking misses command/builtin/backslash/prefix-assigned cd and pushd


cameronsjo:
`guard-rm`'s `cd` tracking matches the literal token `cd`, so every other spelling of a directory change leaves the effective directory stale and a relative delete target is judged against a directory the shell has already left.

## Reproduction

Verified against a build of `main`, all from `cwd=/private/tmp`:

| Command | Verdict | Should be |
|---|---|---|
| `cd ~; rm -rf Documents` | BLOCK | BLOCK |
| `command cd ~; rm -rf Documents` | **ALLOW** | BLOCK |
| `builtin cd ~; rm -rf Documents` | **ALLOW** | BLOCK |
| `\cd ~; rm -rf Documents` | **ALLOW** | BLOCK |
| `FOO=1 cd ~; rm -rf Documents` | **ALLOW** | BLOCK |
| `pushd ~; rm -rf Documents` | **ALLOW** | BLOCK |

`bash` confirms all five move `$PWD`. The plain `cd` form blocks correctly.

## Root cause

`crates/guardrails/src/guard_rm.rs`, in `collect_targets`:

```rust
if tokens.first().map(String::as_str) == Some("cd") {
```

This reads the **raw** token list — not `argv` after `skip_transparent_prefixes`, and without the `command_word` basename/backslash-strip that the delete-verb check just below it uses. `pushd` is not modeled at all.

The delete verb two branches down goes through `command_word(first)` on the post-prefix `argv`, so `\rm` and `/bin/rm` are both caught. The `cd` arm predates that and never got the same treatment.

## Suggested shape

Route the `cd` check through the same `command_word(argv.first())` path the verb check uses, and add `pushd` — resolving its target the same way, since `pushd <dir>` changes the working directory exactly as `cd` does. `popd` has no argument to resolve and should set the directory unknown.

Note there is now a `dir_known` flag in this function (added alongside the variable-expansion work) whose whole purpose is to mark the directory unresolvable rather than carry a stale one; these spellings should feed it.

Found by an adversarial security review of a separate guard-rm change (cameronsjo/claude-configurations#361 cluster F). Not introduced by that change; present on `main`.

cameronsjo:
## Attempted three ways across five adversarial review passes, and withdrawn

[...]

PR #658, opened 2026-08-08
fix(guard-rm): protections outrank the temp carve-out; an empty substitution no longer downgrades Block

### LINKED ISSUE #541: guard-rm: a trailing empty substitution downgrades rm -rf ~/Documents from Block to Ask
## Summary

Appending an empty command substitution to a destructive `rm` downgrades `guard-rm` from **Block** to **Ask**, on a command `bash` executes identically to the blocked form.

| Input | verdict | what bash does |
|---|---|---|
| `rm -rf ~/Documents` | **Block** | deletes |
| ``rm -rf ~/Documents`` `` (empty backticks) | Ask | runs as plain `rm -rf ~/Documents` |
| `rm -rf ~/Documents$()` | Ask | same |
| `rm -rf ~/Documents${EMPTY}` | Ask | same |
| ``rm -rf ~/Documents`true` `` | Ask | same |

Measured against a built binary on both `origin/main` and a feature branch — identical on each, so this is pre-existing and not a regression from any in-flight work.

## Why it is a downgrade rather than a bypass

`Ask` is the fail-safe verdict, and it is doing its job: the target resolves to `Unresolvable`, so the guard declines to judge rather than guessing. Under a harness that converts `Ask` to `Block` this costs nothing. Under one that surfaces `Ask` as a prompt, the protection for the single most destructive shape the guard exists to stop drops from "refused" to "confirm?" — and the operator confirming a prompt they have seen many times is exactly the erosion the Block verdict was chosen to avoid.

The trigger is a suffix that changes nothing about what runs. That is what makes it worth closing rather than documenting: there is no legitimate reason for a trailing empty substitution to alter the verdict, and the delta between the two rows is invisible to a reader skimming the command.

## Where this came from, and the claim it corrects

Found while reviewing #509's branch. That branch's new test and CHANGELOG assert the reported case "already reaches its fail-safe `Ask` verdict". That is true of the input actually sampled — ``rm -rf ~/Documents` `` with a single unmatched backtick — but that input is a **bash syntax error**, so nothing runs and `Ask` costs nothing. The claim was then generalized to the shape as a class. Its executable siblings, above, are not syntax errors.

So the disposition is sound for what was measured and overstated for what was implied — a good illustration of a sampled input standing in for a class.

## Suggested fix

Resolve the target after removing substitutions that provably expand to nothing, so the four rows above resolve the same way `rm -rf ~/Documents` does and reach the same Block.

Direction check: this makes a detector stricter, which per the invariant in `command_word`'s doc comment can only add blocks. The control that must not move is a substitution whose expansion is genuinely unknown — that must keep resolving to `Unresolvable`/`Ask`, because guessing there is the failure mode `Ask` exists to prevent. `` rm -rf ~/`cat which_dir` `` is the shape to pin as still-Ask.

## Wanted before merge

A differential over the five rows above plus at least one unknown-expansion control, driven through the real check rather than the helper — the same reason the branch that surfaced this had a green 1,141-test suite and no coverage of it.

### LINKED ISSUE #569: guard-rm: a $TMPDIR pointing at home (or any real tree) silently disables protection for that whole subtree
`guard-rm` resolves its temp roots partly from `$TMPDIR` (`crates/guardrails/src/guard_rm.rs`, temp-root check). A `$TMPDIR` pointing at a directory that contains real work silently reclassifies that entire subtree as disposable, and the protections that would otherwise fire on it stop firing — with no message, no ledger row, and no other signal.

## Reproduction

Piping a `PreToolUse` payload straight to the hook (the `try` subcommand substitutes its own cwd, so use the bare subcommand):

```
command      : rm -rf /Users/<user>/Documents
TMPDIR=/var/folders/xx/T/   -> BLOCK  "this deletes a top-level entry in your home directory"
TMPDIR=/Users/<user>        -> silent ALLOW
```

`TMPDIR=/` still blocks, so the root case is handled; it is the "TMPDIR set to a directory that actually contains things" case that opens.

Verified on 0.70.1.

## Why this is worth fixing even under an accidents-not-adversaries threat model

The obvious framing is an attacker setting `TMPDIR`, and under this guard's stated threat model that is out of scope. The accident case is the real one and does not need an adversary:

- `TMPDIR` is settable in `.claude/settings.json`'s `env` block, so a well-meaning "put temp files somewhere with more space" edit pointed at a project or home directory disables deletion protection for that whole tree, permanently and invisibly.
- Build systems, sandboxes, and CI wrappers routinely re-point `TMPDIR` at a workspace-local directory. A session inheriting one of those loses protection for exactly the tree it is working in — the tree that matters most.

The failure is silent in the direction that costs data. There is no nudge, and `Outcome::Allow` writes nothing to `denials.jsonl`, so nothing distinguishes "classified and allowed" from "protection was off".

## Suggested shape

Constrain what `$TMPDIR` is allowed to contribute rather than trusting it wholesale. Options, roughly in order of preference:

1. Ignore a `$TMPDIR` that is an ancestor of, or equal to, the home directory (and, more generally, one that contains any protected root). A temp root that swallows a protected root is not a temp root.
2. Evaluate the protected classifications (home, vault, git repo) *before* the temp carve-out, so an explicit protection always wins over a disposability claim.
3. At minimum, require `$TMPDIR` to be under a known system temp prefix before honoring it.

Option 2 has a second benefit worth noting: the current ordering also means `rm -rf /tmp/<dir-containing-.git>` is a silent Allow, because Temp is evaluated before the git-repo check. That matters here because this estate deliberately parks real checkouts under temp (`git worktree add /tmp/<wt>` appears in its own operating docs), so a worktree with uncommitted work is deletable with no prompt.

## Detection already landed

`guard-rm-liveness` (cameronsjo/cadence-hooks#568) carries a home-directory Block probe specifically for this: it runs under the session's real `$TMPDIR`, so a widened one makes the probe flip Block to Allow and the check nudges. The three other probes all still pass in that state — none of their operands falls under the widened root — which is why the home probe was added. That is detection, not a fix; this issue is the fix.

### LINKED ISSUE #576: guard-rm: Temp classification pre-empts the git-repo check, so a repo under /tmp deletes silently
`guard-rm` evaluates the Temp classification **before** the git-repo check, so a git repository living under a temp root is a silent Allow:

```
rm -rf /tmp/<dir-containing-.git>          -> silent ALLOW
rm -rf ~/Projects/<dir-containing-.git>    -> BLOCK "this deletes a git repository"
```

Verified on 0.70.1 with an otherwise identical fixture in each location — the only difference is the parent directory.

## Why this is not merely theoretical here

This estate deliberately parks real checkouts under temp. Its own operating docs prescribe it: *"A plugin checkout parked on a peer's unpushed feature branch is workable without touching it: `git -C <plugin> worktree add /tmp/<wt> -b <branch> origin/main`"*. A worktree created that way holds real, often uncommitted work, and `rm -rf` on it carries no prompt at all.

The design intent behind the temp carve-out is sound — temp roots are disposable, and prompting on scratch deletes is exactly the friction that got the blanket `ask` rows retired. The problem is that "disposable by location" and "contains a git repository" are both true of these paths, and the current ordering lets location win silently.

## Suggested shape

Evaluate the protected classifications (git repo, vault, home) **before** the temp carve-out, so an explicit protection always beats a disposability claim. A `.git` under `/tmp` is then a normal Block, and genuinely scratch deletes under temp are unaffected because they have no `.git`.

Cost: deletes of throwaway repos under `/tmp` — test fixtures, this project's own `Scratch` git fixtures — would start prompting. Worth checking whether the fixture paths in `crates/core/src/git_fixtures.rs` would trip it, and whether the existing `.claude/worktrees/` scratch carve-out should stay ahead of the git check even after the reorder.

## Relationship to the other TMPDIR issue

Split out of cameronsjo/cadence-hooks#569, where this appeared only as a closing paragraph. They share a root — an ordering that lets a temp claim pre-empt a protection — and #569's suggested option 2 is the same reorder. Separate issues because they have different triggers and different blast radii: #569 needs a widened `$TMPDIR`, this one needs nothing but a checkout in the default temp root.

cameronsjo:
Three guard-rm verdict-widening fixes, one PR — every change moves verdicts toward Block, never toward Allow.

- **#541** — a lexically-empty substitution (`$()`, empty backticks) no longer downgrades Block to Ask: `strip_empty_substitutions` removes provably-empty bodies before classification, with rails (stripping-to-empty stays Unresolvable; non-empty bodies keep their Ask verbatim). `${EMPTY}` and `` `true` `` deliberately stay Ask — resolving them means modeling variable/command values, the widening class that was rejected three times on #428.
- **#569** — a `$TMPDIR` that equals or covers `$HOME` no longer grants the temp carve-out. The covers-home veto gates the whole `via_env` branch on both the raw and canonicalized values (security review caught the raw-branch gap: a symlinked `TMPDIR=/scratch -> $HOME` or case-differing `/users/dev` on APFS previously still swallowed home). Shared-predicate blast radius is in the safe direction (more blocks in enforce-worktree; fixture relocation covered by the suite).
- **#576** — a git repo under a temp root now Blocks: explicit protection outranks the location claim, guard-locally (pathclass precedence untouched); `.claude/worktrees/` disposability still wins. The stale liveness doc-comments describing the old ordering are corrected in the same diff.

Security review ran on the diff (independent adversarial pass): one Important finding (the raw-branch veto gap above) — fixed in `be0ec14`; the `.GIT` case-fold nit filed separately as cameronsjo/cadence-hooks#657. No Cargo.toml bump — the 0.76.0 chore PR ships after the batch.

Known machine-local reds excluded from the gate (fail identically at clean origin/main): pending_wiring_hooks_are_still_unwired, all_binary_subcommands_are_registered.

Closes cameronsjo/cadence-hooks#576
Closes cameronsjo/cadence-hooks#569
Closes cameronsjo/cadence-hooks#541

Session-Name: tidal-ballad
Session-Id: 6bff6f72-e4d7-4e6f-9e7a-8df28d520b1f
Model: claude-fable-5
Harness: claude-code 2.1.226
Machine: cf6e768835c7

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,12 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
 ## [Unreleased]
 
+### Fixed
+
+- **`guard-rm` no longer treats a protected target as unresolvable because an empty substitution was appended (#541).** `$()`, `$( )`, and an empty backtick pair expand to nothing in every shell, so `rm -rf ~/Documents$()` names exactly `~/Documents` — but the operand reached the classifier still carrying a `$` and downgraded a contractual Block to an Ask, a downgrade bought with two characters. Empty substitutions are now removed before any other parsing touches the segment, which also covers the two shapes the shell layer mangled: a trailing `$()` whose `)` the group-wrapper trim consumed, and `$( )` split across two tokens. Only *lexically* empty bodies are removed — `$(cat f)` and `${EMPTY}` keep their Ask verbatim, and an operand that was all empty substitution still asks rather than falling through to the working directory's verdict.
+- **A `$TMPDIR` that is the home directory, or an ancestor of it, is no longer honored as a temp root (#569).** Pointing `$TMPDIR` at `$HOME` reclassified every path under home as disposable scratch, making `rm -rf ~/Documents` a silent allow. The value is now refused across `guard-rm`, `enforce-worktree`, and the shared path classifier — component-wise, and under either spelling: a `$TMPDIR` that only *resolves* to home is rejected too, closing a symlinked `$TMPDIR` and a `$TMPDIR` differing from `$HOME` only in case on a case-insensitive volume. Real temp roots are unaffected, including macOS `/var/folders` and the Windows per-user temp directory that legitimately lives *under* the profile directory.
+- **An explicit `.git` now outranks `guard-rm`'s temp carve-out (#576).** A repository checked out under `/tmp` is still a repository, so `rm -rf /tmp/repo` blocks instead of allowing on the strength of its location alone. cadence's own managed scratch keeps its allow: a `.claude/worktrees/` checkout is disposable by construction, including when it lives under a temp root.
+
 ## [0.75.0] - 2026-08-08
 
 ### Added
--- a/crates/core/src/git_fixtures.rs
+++ b/crates/core/src/git_fixtures.rs
@@ -101,6 +101,7 @@ pub(crate) fn resolve_scratch_dir(root: &Path, tag: &str) -> PathBuf {
         std::env::var("TMPDIR").ok().as_deref(),
         override_root.as_deref(),
         xdg_cache_home().as_deref(),
+        Some(&crate::paths::user_home_lossy_or_default()),
     )
 }
 
@@ -138,13 +139,14 @@ fn resolve_scratch_dir_with(
     tmpdir: Option<&str>,
     override_root: Option<&Path>,
     xdg_cache_home: Option<&Path>,
+    home: Option<&str>,
 ) -> PathBuf {
     // `$TMPDIR` is invariant for the process — canonicalize once (a real
     // `realpath` syscall chain) and reuse it across all three candidates,
     // rather than re-canonicalizing per candidate inside `escapes_carveout`
     // (cadence-hooks#403 code review: the relocation path this function
     // exists for is precisely the one that was paying that cost 2-3x).
-    let tmpdir_canonical = crate::worktree::canonicalize_tmpdir(tmpdir);
+    let tmpdir_canonical = crate::worktree::canonicalize_tmpdir(tmpdir, home);
     let suffix = format!("{tag}-{}", std::process::id());
     let candidate = |base: &Path| base.join(&suffix);
     // Canonicalize `dir` (or its nearest existing ancestor — the `{tag}-{pid}`
@@ -162,7 +164,12 @@ fn resolve_scratch_dir_with(
     let escapes_carveout = |dir: &Path| {
         let resolved = canonicalize_nearest_existing(dir);
         !is_claude_managed_dir(&resolved)
-            && !path_under_temp_root_with_canonical(&resolved, tmpdir, tmpdir_canonical.as_deref())
+            && !path_under_temp_root_with_canonical(
+                &resolved,
+                tmpdir,
+                tmpdir_canonical.as_deref(),
+                home,
+            )
     };
 
     let default_dir = candidate(root);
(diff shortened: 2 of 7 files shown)


cameronsjo:
Gate record: CodeRabbit shows green with no review object (estate-wide rate-limit exhaustion today). Merge gate of record: an independent Opus security review ran against the diff (one Important — the raw-branch $TMPDIR veto gap — fixed in be0ec14 and re-verified at source; .GIT case-fold nit filed as #657), and the orchestrator ran the full workspace suite (--workspace --no-fail-fast) confirming all 16 new named tests ok; only the two machine-local audit reds failed, which fail identically at clean origin/main. This merge trusted the internal review chain, not CodeRabbit.

PR #708, opened 2026-08-16
docs: release recipe's direct-push-to-main is stale under the estate ruleset

cameronsjo:
Discovered shipping 0.79.0 this session: the documented `git push origin release/X:main` recipe fails `GH013: Changes must be made through a pull request` — the estate-wide branch-protection ruleset now requires PRs everywhere, even for a clean fast-forward. Updates the recipe to push the branch + open a PR + merge (auto-tag still fires on the merge).


cameronsjo:
Closing as superseded by #772.

**The diagnosis was right.** The direct `git push origin release/X:main` fast-forward is genuinely dead under the estate ruleset — it bit the 0.88.0 release today, which is what sent me back to this PR.

**But `main` already carries the fix.** #772 landed the same correction on 2026-08-26 as `038c579`, and its wording is better sourced — it cites the 0.86.0 release as the verification rather than a general observation:

> Do the bump in a worktree **and land it via a PR** … push the branch, and open + merge a `chore(release): X` PR (the merge to main triggers auto-tag). The old direct `git push origin release/X:main` fast-forward is **DEAD** — the estate ruleset rejects any direct push to main (`push declined due to repository rule violations`, verified on the 0.86.0 release, 2026-08-25).

**Merging this now would be a regression.** The branch is 82 commits behind, and its replacement paragraph still names "the explicit-`--workspace` report regeneration from step 2 above". That step no longer exists — #753 retired the Codex compatibility report along with `scripts/generate-codex-report.py`, the `report`/`report-check` Makefile targets, and the CI freshness gate, dropping the recipe from four steps to three. Merging would put a dead instruction back into the release recipe pointing at a step that isn't there. `git merge-tree` against the merge-base also reports a conflict on that exact paragraph, so it could not fast-forward regardless.

The only thing here that `main` lacks is the `GH013: Changes must be made through a pull request` string, where `main` quotes `push declined due to repository rule violations` — two lines of the same rejection output, so not worth a follow-up.

PR #719, opened 2026-08-18
fix(core): surface the tail behind a backtick span's unterminated quote

### LINKED ISSUE #653: shell: backtick substitution arm misses the unterminated-quote tail — the #551 $( ) fix was not applied to backticks
## What

The backtick arm of `substitution_bodies` (`crates/core/src/shell.rs:2584`) has the exact unterminated-quote-swallows-the-tail bug #551 fixed for `$( )`:

```
echo `echo '` && cat .env
```

Live bash: the substitution errors on the unterminated quote, `echo` still succeeds, and `&& cat .env` runs. Parser: `cat .env` never becomes a segment; ALLOW on pre- and post-#551 binaries alike.

The arm's doc comment cites `substitution_spans_backtick_closes_even_inside_single_quotes` to justify keeping backticks quote-untracked — correct for the *truncation* question, but it does not reach this *unterminated-quote tail* question, where #551's both-readings answer for `$( )` was not applied.

## Remediation

Apply the same both-readings emission to the backtick arm when the span's content carries unbalanced quoting. No change to backtick span semantics — the existing truncation test stays green; only an additional body is emitted alongside.

Found by the adversarial security review over #551. Sibling of the confident-terminator issue filed alongside.

cameronsjo:
The backtick arm of the shell parser collected its span quote-blind and emitted only the body — correct for truncation, but the span's unmatched quote swallowed the trailing `&& cat .env` from segmentation, so no guard ever saw it (live repro in the issue). Mirrors the `$( )` arm's both-readings emission from #656: when the span's quoting is unterminated, push the body AND the tail. Additive-only — a balanced-quote control confirms no spurious tail emission.

Closes #653

Security note: guard-feeding parser logic — an independent adversarial security review runs at the merge gate before this lands.

```text
Model: claude-sonnet-5 (impl-653b Stage B; Fable orchestrator)
Harness: claude-code 2.1.234
Machine: cf6e768835c7
```

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved command parsing when backtick substitutions contain unterminated single or double quotes.
  - Ensured trailing commands are still detected and blocked when they attempt to read `.env` files.
  - Preserved correct handling for balanced and escaped quotes.

- **Tests**
  - Added regression coverage for malformed quoting and secret-leak prevention scenarios.


cameronsjo:
Gate-2 adversarial security review (independent Opus pass on the guard-feeding parser change): **CLEAN**.

- No false-allow regression: the guard consumes via `expand_segments`, which recurses on every emitted body, so the pushed whole-tail re-surfaces sibling commands (`&& cat .env` → its own segment → Block). Root safety: `span_quoting_unterminated` and `split_segments` share quote/escape semantics, so the tail reaches the guard whether the helper returns true (surfaced here) or false (segmented normally) — a miss would require the two readers to disagree, which shared semantics prevent. The change captures a **superset** of the old emission.
- Additive-only confirmed: balanced/escaped-quote controls assert zero tail bodies; the new push fires strictly on the unterminated case.
- No panic/OOB on the slice ops (bounds traced for empty span, backtick-at-end, unicode via `Vec<char>`).
- Tests: 268 (core/shell) + 233 (guard) + the 4 new controls + the guard regression, all pass.

Polish marker re-recorded `security=ran` on the shipped tree. This clears the merge-gate security blocker noted at PR open.

PR #723, opened 2026-08-18
fix(persist-plan): recognize persisted plans across filenames, sessions, and trigger paths

### LINKED ISSUE #695: persist-plan: make the row-tier skip observable — a forged plan-links row suppresses persists silently
## Context

The #690 fix added a row-keyed tier-0 idempotency check to `persist-plan`'s late-scan path: a `plan-links.jsonl` row matching (`body_sha256`, `child_session_id`) makes the turn a **silent** `allow` — deliberately, since the row is only ever written by a persist that already nudged once, and the per-prompt re-nudge was the noise #690 removed.

The independent security pass on that diff raised one Important finding worth its own ruling:

## Finding

A **forged or corrupted row** breaks the "a row implies a prior nudge" assumption, and nothing observes the break:

- The row tier short-circuits **before** the dir scan, so the dir scan backstops a *missing* row, not a *false* one. The fail-open story (ADR-0001) holds for absence and malformation; a present-but-false row fails **closed on the persist**.
- The skip is byte-indistinguishable from an ordinary quiet turn. The operator's only signal that persistence works — the nudge — is exactly what the tier suppresses.
- Realistic actor: not remote — the agent itself under prompt injection, or any same-user process. Appending one JSONL line to the machine-local metrics file is materially cheaper than forging a `docs/plans/*.md` with a matching frontmatter hash (what defeating the dir scan required).

Net impact is loss of a provenance artifact (the approved plan never lands in `docs/plans/`), not code execution or exfiltration.

## Suggested remediation

Make the row-tier skip **observable, not louder**: emit one debug/metrics record (e.g., a row in an existing local telemetry stream) naming the matched row's `plan_path` when the tier suppresses — so a phantom row is detectable after the fact. Explicitly **not**: gating the skip on `plan_path` still existing (that would defeat the delete-and-retarget workflow the tier was built for), or re-nudging (the #690 noise).

## Also on the record

Tail-window flushing (~256 KiB of appended noise pushes real rows out) degrades only idempotency, never safety — documented and tested (`row_beyond_the_tail_window_fails_open_to_the_dir_tier`); no action needed.

### LINKED ISSUE #696: feedback(persist-plan): hand-persisted plans are invisible to body-hash idempotency — late scan re-persists under its own slug
## What happened

Post-#693 (v0.77.0, binary verified current at write time), persist-plan's late scan re-persisted an approved plan that the executing session had **already persisted by hand and committed** — producing a duplicate under a hook-derived slug (`2026-08-14-editorial-review-recent-plan-docs-structure-prose.md` vs the committed `2026-08-14-editorial-review-recent-plan-docs.md`, `claude-configurations@28a5600`). Full timeline + source trace: cameronsjo/cadence-hooks#690 (comments from session copper-chisel, 2026-08-14).

## Why

Idempotency keys on `body_sha256` at both tiers — tier 1 scans `docs/plans/` for a doc carrying the hash (`crates/session/src/persist_plan.rs:734` @ v0.77.0), tier 2 checks `(body_sha256, child_session_id)` rows (`session_already_persisted`, `:548`). A hand-persisted plan (using-cadence's persist-it-now path, which the skill mandates when no hook context line appeared) carries no `body_sha256:` frontmatter and its body has drifted from the approved-plan text (producer tuple, Panel stanza, checkbox ticks, deviations). No shared key → the late scan sees an unpersisted approval.

## Proposed fix

The late scan additionally treats the approval as covered when any plan doc in `docs/plans/` carries `approved_session_id:` equal to the approval it is about to persist — robust to every rendering divergence, no skill change required. Alternative (declined-leaning): have using-cadence stamp `body_sha256` of the approved body into manual persists — works, but every manual writer must know the hook's hashing rules, and one missed stamp reproduces the dup.

Session-Name: copper-chisel
Session-Id: 30256786-4d94-4803-9b07-fdf282afb69d
Model: claude-fable-5
Harness: claude-code 2.1.233
Machine: cf6e768835c7

### LINKED ISSUE #697: feedback(cadence-hooks): session persist-plan re-persists a plan another session already committed
## Problem

`session persist-plan` re-persisted an approved plan that a *different* session had already persisted **and committed**, producing two plan docs for one workstream under two different slugs. The v0.77.0 idempotency fix (#690) did not catch it — that skip is same-session/same-row scoped, and this is the cross-session case.

## Environment

- `cadence-hooks 0.77.0` (carries the #690 re-persist fix; verified `--version`)
- Repo: `cameronsjo/claude-configurations`, branch `main`, shared checkout with live peers
- Sessions: `iron-chisel` (`85ce11e3-592b-463c-8aa4-ea021e19af50`) receiving; plan approved in `gilded-lathe` (`8df53d4c-7459-46a8-81d5-39b01a1c04ff`)

## What happened

`iron-chisel` was handed an approved plan as its opening prompt, wrote it to `docs/plans/2026-08-14-cli-capture-ctrl-space-leader-and-capture-launcher.md`, committed it, and executed all ten steps to completion (two commits, an upstream PR).

Nine turns later, mid-execution, the `UserPromptSubmit` hook fired and wrote a **second** file for the same approved plan:

```
docs/plans/2026-08-14-cli-capture-launcher-ctrl-space-leader-a-cadence-research-sc.md
```

with the context line:

> Approved plan persisted to … (approved in gilded-lathe). Verify placement, then commit it (explicit-path git add) before implementation.

— instructing a second index be committed for work that was already done.

Its frontmatter is internally correct (`session: iron-chisel` / `approved_in: gilded-lathe`), so the attribution is not the bug. The bug is that nothing asked whether a plan doc for this workstream already existed in `docs/plans/`.

## Second observation, same fire

The same context line carried three format gates against the file the hook had **just written**:

> panel gate: this plan carries no settled `Panel:` line … format gate: no Alternatives-declined stanza … format gate: no checkbox tasks

The hook persists raw plan-mode text, which structurally cannot carry a `Panel:` line, an Alternatives stanza, or `- [ ]` checkboxes. So every hook-persisted plan is born non-conformant and immediately flagged, and the remediation burden lands on whichever session happens to receive it. Filing as one issue since both come from the same fire; split if you'd rather.

## Expected vs actual

- **Expected:** skip (a committed plan doc for this approved plan already exists on this branch), or persist and say it supersedes the existing doc.
- **Actual:** wrote a duplicate under a different slug and instructed the session to commit it.

## Evidence

The duplicate was never tracked and never committed (`git log --all -- <path>` empty), carried zero checkboxes / `Panel:` / Alternatives / Deviations markers (`grep -cE` → 0), and was 122 lines of plan prose against the 75-line executing copy. Deleted rather than committed; recorded in the surviving plan's `## Deviations`.

**Not reproduced from a synthesized payload** — I have the observed fire and the artifact on disk, not a rebuilt `UserPromptSubmit` payload that re-triggers it. The wiped-approval-turn precondition is the part I could not synthesize cheaply.

## Possibly relevant

#690 / #695 are the same subsystem. Whether the dedupe key can reasonably reach across sessions — or whether the answer is a `docs/plans/` scan on the branch rather than a ledger row — is the maintainer's call.

### LINKED ISSUE #703: persist-plan writes a duplicate plan doc on every turn when the plan is already persisted under a different filename
## Problem

`session persist-plan` writes a **new** plan document every time it fires, without checking whether that plan is already persisted in `docs/plans/`. In a session executing an approved plan, it fired on **every** `UserPromptSubmit` turn — re-creating its copy after each deletion — even though the plan had already been persisted, committed, and reconciled to `status: done` with every checkbox ticked.

Three properties compound it:

1. The output filename is derived from the plan's H1 title plus the **current date**, so a plan persisted under any other filename is invisible to the check, and a multi-day session produces a *new* duplicate each day rather than converging on one file.
2. The written frontmatter is `status: "in-flight"` unconditionally — so the duplicate outranks a committed `status: done` plan in any in-flight scan — including the one the end-of-session cleanup flow performs when it looks for a living plan to reconcile.
3. The body written is whatever the prompt carried. For the approve-plan → `/clear` → execute workflow the prompt is the **pre-execution** snapshot, so the duplicate is frozen at the plan's approved state while the real plan advances — unticked boxes, no `## Deviations`.

Net effect is two plan documents for one workstream, the stale one self-describing as in-flight. That is the index-staleness hazard the plan-execution rules warn about, arriving from the tooling rather than from drift.

## Diagnosis

`cadence-hooks 0.78.0`.

The sample payload does not reproduce it — `cadence-hooks try session persist-plan` uses `cwd: /nonexistent-cadence-hooks-try-sandbox` and exits ALLOW with no output. Rebuilding the real payload does reproduce it.

Payload: a `UserPromptSubmit` whose `prompt` is `"Implement the following plan:\n\n" + <plan body>` and whose `cwd` is a repo that **already contains that exact plan**, committed, at `docs/plans/<descriptive-slug>.md` with `status: done` and 21/21 boxes ticked.

```
$ cadence-hooks try session persist-plan --payload /tmp/pp-real.json

Outcome:  NUDGE (exit 0)
Context injected (what Claude sees):

  Approved plan persisted to <repo>/docs/plans/2026-08-15-<title-derived-slug>.md
  (approved in unknown). Verify placement, then commit it (explicit-path git add)
  before implementation.
```

A second plan file appears in `docs/plans/`. Diffed against the already-committed one, the bodies are identical (I fed it the current body) — the divergence is the filename, the date prefix, and `status: "in-flight"` vs `status: done`.

Note the two files also differ in date: the committed plan is dated the day it was approved, the new one carries today's date, so they will never collide into a single path.

## Root Cause

Confident on the observable behavior, uncertain on the internals: the persist decision appears to key on the *prompt shape* (an approved-plan-looking prompt) and to resolve its destination path from the title + date, with no lookup of existing plans in `docs/plans/` and no check of an existing plan's `status:`.

## Suggested Fix

Reporting the seam rather than designing the fix, but the two checks that would have caught this session:

- **Match on identity, not path.** The plan's `approved_session_id` frontmatter key already exists for exactly this — recognizing the same plan across a `/clear` boundary. Scanning `docs/plans/` for a plan carrying the same `approved_session_id` (or the same `body_sha256`, which this hook already computes and writes) would find the existing file regardless of what it is named.
- **Stand down on a terminal status.** A plan at `status: done` has nothing left to persist; re-persisting it as `in-flight` actively corrupts the in-flight scan that other cadence surfaces depend on.

Confidence: the reproduction is confirmed; the fix direction is a suggestion.

## Related

There is a second, adjacent persister — `session persist-plan-approval` (`PostToolUse:ExitPlanMode`). This machine additionally had a *local* shell hook doing the same job, i.e. three mechanisms persisting one plan. The local one has been unregistered. Whether the two shipped persisters should coordinate (or whether one subsumes the other) is a separate question from the duplicate-write defect above, and I have not investigated it.

## Labels

cadence-hooks, bug

### LINKED ISSUE #715: feedback(persist-plan): full three-gate block fires on an outro message for an already-executed plan
Observed 2026-08-17 (session fern-chisel, artificer-voice work): the plan was approved in a predecessor session, executed to completion, and both work commits were already pushed. On the user's next message ("Let's outro"), the persist-plan UserPromptSubmit hook emitted its full context block: persistence line plus panel gate, Alternatives-declined gate, and checkbox gate — roughly 1k chars of instruction, all pointed at pre-implementation steps ("run the plan-review panel before implementing") for work that had already shipped.

The persistence itself was correct and useful (the plan file landed on disk and was committed). The friction is the gate block's size and timing:

- The gates assume implementation is ahead; they have no way to be right on a post-execution turn, and the instructions they inject are mistimed.
- The block lands on every subsequent user message until the plan is committed, so a terse outro message pays the full ~1k chars.

Ask: keep the persistence line, make the gate block terse — one line naming the missing stanzas ("plan lacks Panel/alternatives/checklist — see plan template") rather than the full instruction text. The template already carries the how.

Session-Name: fern-chisel
Session-Id: 7a653180-b992-4ed2-be46-875034148c25
Model: claude-fable-5
Harness: claude-code 2.1.233
Machine: cf6e768835c7

cameronsjo:
The persist-plan hook re-persisted already-committed plans — a duplicate doc every turn (#703), invisibly (#695), blind to hand-persisted plans (#696), across sessions (#697), behind a ~1k-char triple gate (#715). Five fixes on one branch:

- T1 — both recognition layers (row tier + dir hash scan) now apply to the prefix trigger path, not just late-scan.
- T2 — a third layer matches an on-disk plan's `approved_session_id:` frontmatter (exact compare, untrusted-input safe) and appends a plan-links row, giving the machine cross-repo memory. This is the fix for today's live wrong-repo re-fire.
- T3 — row-tier skips are now observable via a fail-open, capped `plan-skips.jsonl`.
- T4 — the three full-instruction gate blocks collapse to one terse line naming only the missing stanzas.

Verified against HEAD: #697's via-hook face was already fixed at v0.78.0; its hand-persisted face and the other four are fixed here.

Closes #703, closes #695, closes #696, closes #697, closes #715

Security note: T2's `approved_session_id` predicate drives persist-suppression — an independent adversarial security review runs at the merge gate before this lands (same forgeable-frontmatter blast radius as the existing hash tier; on the record).

```text
Model: claude-sonnet-5 (impl-703b Stage B; Fable orchestrator)
Harness: claude-code 2.1.234
Machine: cf6e768835c7
```


cameronsjo:
Gate-2 adversarial security review (independent Opus pass on the T2 `approved_session_id` predicate): **CLEAN**.

- Injection: the frontmatter value reaches exactly one operation — an equality compare against the in-memory resolved parent session id. Never rendered, interpolated, logged, or used to build a path/command/flag. `yaml_unquote` is a pure string transform. No injection reachable.
- Suppression blast radius: a forged `approved_session_id:` is the **same accepted forgeability class** as the existing `body_sha256:` hash tier — and narrower (harder to forge, and layer 3 only fires after a hash-miss *and* a real parent resolves). Worst case is one suppressed persist; `CADENCE_PERSIST_PLAN_FORCE` overrides. Not a new class.
- Path handling, fail-open writer, tests (129 pass): all clean.

Polish marker re-recorded `security=ran` on the shipped tree. This clears the merge-gate security blocker noted at PR open.

PR #725, opened 2026-08-18
fix(guard-rm): case-fold the .git component match — .GIT no longer evades GitRoot

### LINKED ISSUE #657: guard-rm: .git component match is byte-exact — /tmp/repo/.GIT evades GitRoot on case-insensitive volumes

cameronsjo:
On a case-insensitive volume (APFS default), `rm -rf /tmp/repo/.GIT` deletes the real `.git` — but the guard's component match was byte-exact (`seg == ".git"`), so the path classified as `Temp` and took the temp carve-out's ALLOW. Both mirrored match sites now use `eq_ignore_ascii_case`.

The fold is deliberately `.git`-only. Sibling segment compares (`.claude`, docs/plans, temp, home, vault) stay byte-exact because they feed allow-side carve-outs, where folding would *widen* a silent ALLOW — the wrong direction. The disk probe `is_git_root_on_disk` needs no change: `stat` already folds case on case-insensitive volumes.

Closes #657

Note: guard-rm is currently disabled on the operator's machines (handed to auto-mode policy), so this changes no live fleet behavior until re-enabled — the guard still ships and the fix is still owed.

Security note: deletion-guard logic — an independent adversarial security review runs at the merge gate before this lands.

```text
Model: claude-opus-5 (impl-657 Stage B; Fable orchestrator)
Harness: claude-code 2.1.234
Machine: cf6e768835c7
```


cameronsjo:
Gate-2 adversarial security review (independent Opus pass): **1 finding, Low — ship it.**

The reviewer measured a real before/after delta (probe test run on the branch and again with the fold reverted) rather than reading the code:

```text
                     before  ->  after
rm -rf /a/repo/.GIT           Ask    -> Block
rm /a/repo/.GIT/config        Allow  -> Block
rm /home/.claude/.GIT/*       Ask    -> Block
rm /a/repo/.GIT/objects/*     Ask    -> Block
```

- **Direction**: all three `has_git_component` call sites fold protectively — including the negated one in the symlink-demotion guard, where folding *shrinks* the demotion's domain.
- **Scope**: only the two `.git` compares changed; `.claude`, docs/plans, temp, home, and vault stay byte-exact (folding the allow-side ones would widen a silent ALLOW).
- **Unicode**: `eq_ignore_ascii_case` is sufficient — no Unicode string APFS resolves to `.git` escapes the ASCII fold (Kelvin sign, dotless i, dotted capital I, and fullwidth forms all checked).
- **Low finding**: a flat glob inside a `.GIT`-spelled directory softens `Ask`→`Allow` via the existing repo-artifact-sweep arm. On APFS that directory *is* `.git`, which already allows it by design — the fold makes two spellings agree. Not a new class.
- Tests: 35 pathclass + 135 guard_rm pass.

A pre-existing residual it surfaced (`~/.CLAUDE/...` Allows where `~/.claude/...` Asks) is out of this diff's scope and filed as #726.

Polish marker recorded `security=ran` on the shipped tree.

PR #731, opened 2026-08-19
fix(guard-rm): case-fold under_claude_dir — .CLAUDE no longer takes a silent Allow

### LINKED ISSUE #726: guard-rm: under_claude_dir stays byte-exact, so ~/.CLAUDE/... silently Allows where ~/.claude/... Asks
## Finding

Surfaced by the independent security review of #657 (PR #725), which deliberately folded **only** the two `.git` segment compares. `under_claude_dir` (`crates/core/src/pathclass.rs`, `*s == ".claude"`) stays byte-exact — correctly, since that diff's scope was `.git`.

Measured on both the #657 branch and its baseline (identical, so pre-existing):

```text
rm ~/.CLAUDE/projects/t.jsonl   -> Allow
rm ~/.claude/projects/t.jsonl   -> Ask
```

On APFS (case-insensitive by default) those are **the same transcript file**. The case variant takes a silent Allow where the canonical spelling prompts.

## Why this is not a one-line fix

The review is explicit that the two `.claude` predicates fold in **opposite** safety directions:

- `under_claude_dir` feeds a **protective** classification (`ClaudeState` → `Ask`). Folding it is protective.
- `under_claude_scratch` feeds an **allow-side carve-out** (the `worktrees`/`intros` allowlist). Folding it would **widen a silent ALLOW** — the wrong direction.

So this wants `under_claude_dir` folded while `under_claude_scratch` stays byte-exact, plus a test asserting that asymmetry so a later "consistency" sweep doesn't fold both. Same reasoning as #657's deliberately `.git`-only scope — worth stating in a comment at both sites.

## Related

- #657 / PR #725 — the `.git` half of this class
- Threat model is accidents-not-adversaries; this is a case-variant typo reaching a silent Allow on transcript state.

_Filed from the independent security review of PR #725 (session bc99ac0f)._

cameronsjo:
Folds the `.claude` path-component compare in `under_claude_dir` to `eq_ignore_ascii_case`, closing the Allow/Ask asymmetry left after #725: `.CLAUDE/...` paths now classify as Claude state (Ask/Block side) instead of falling through to a silent Allow.

The `under_claude_scratch` carve-out deliberately stays **byte-exact** — it is allow-side, and case-folding an allowlist widens it. Tests assert the asymmetry itself in both crates:

- `pathclass::tests::claude_dir_case_variants_are_claude_state`
- `pathclass::tests::claude_scratch_carve_out_stays_byte_exact`
- `guard_rm::tests::claude_state_case_variants_still_ask`
- `guard_rm::tests::claude_scratch_carve_out_does_not_fold_case`

Suites: core 800/800, guardrails 1272/1272 (orchestrator-run, `--no-fail-fast`, rc=0).

Closes #726

Model: claude-fable-5
Harness: claude-code 2.1.234
Machine: cf6e768835c7


cameronsjo:
Gate-2 security review (independent Opus pass, tree `588f168` = current head): **CLEAN** — no Allow-direction path (consulted last in both classifiers; only Source/Unknown displace, protectively), scratch carve-out proven byte-exact through the built binary, no non-ASCII case-fold sibling exists for `claude`'s letters so `eq_ignore_ascii_case` is exact, and the #725 `.git` fold composes cleanly. Pre-existing sibling gap (Home/Vault byte-exact compares) filed separately as #732 rather than widened into this diff.

cameronsjo:
Merge gate: CodeRabbit rate-limited this hour (summary-only, no review object) — gate stood in by the cadence reviewer panel per operator directive: `cadence:code-reviewer` (Sonnet) verdict **MERGE-CLEAN** on head `588f168` (guard table traced independently, all 4 new tests re-run), security floor met by the earlier independent Opus pass (binary-measured). This merge trusts that panel, not CodeRabbit.

